PR #29, opened 2022-07-29
Update `Cargo.toml` format for `cargo component`. 

### LINKED ISSUE #28: Redesign `Cargo.toml` format
Currently, we mix exported and imported interfaces in the dependencies section and that can be quite confusing.

Originally it was intended that the `[package.metadata.component.dependencies]` section would be 1:1 with the normal `[dependencies]` section.

But having to mark some of the entries as `export = true` is clumsy; it's additionally clumsy to treat the default exported interface as the exported interface without a version.

Instead, I propose we change the format to be:

```toml
[package.metadata.component]
# Specifies the name of an export interface to export directly from the component
default = "interface"

[package.metadata.component.imports]
foo = "foo.wit"

[package.metadata.component.exports]
interface = "interface.wit"
bar = "bar.wit"
```

For now I also propose we remove the `version` field since it's currently unused and just use paths to the interface files to make thing cleaner.

peterhuene:
This PR updates the format for `Cargo.toml` to make it clearer as to what
gets imported and what gets exported.

It splits the former `dependencies` table into two tables `imports` and
`exports`.

Each dependency in those tables is uniformly specified as either:

```toml
dep = "path"
```

or:

```toml
dep = { path = "path" }
```

This removes the version numbers from the dependencies for now until such time
as they are useful.

A warning was added if we encounter the old `dependencies` table to warn users
they need to upgrade their `Cargo.toml` format; the tool doesn't attempt to do
so automatically (sorry).

Closes #28.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,8 +190,8 @@ The name of the crate is dependent upon the name specified in `Cargo.toml`:
 ```toml
 # ...
 
-[package.metadata.component.dependencies]
-interface = { path = "interface.wit", export = true }
+[package.metadata.component.exports]
+interface = "interface.wit"
 ```
 
 ## Usage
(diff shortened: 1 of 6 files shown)


radu-matei (approved):
This is great, it makes reading `Cargo.toml` much easier, thanks!

LGTM

README.md
```diff
@@ -212,35 +212,35 @@ More commands will be added over time.
...
 
 To export a _default_ interface (i.e. one where the interface's functions
-are directly exported by the component itself), omit the `version` key:
+are directly exported by the component itself), set the `default` key to
+the name of the entry in the `[package.metadata.component.exports]` table.
 
 ```toml
-binding-name = { path = "path/to/interface.wit", export = true }
+[package.metadata.component]
+default = "foo"
```
guybedford, line 236: I wonder if it would be possible to consider another name like `namespace` or alternatively something more descriptive like `default-interface`? Just to avoid confusing `default` with the `"default"` export name in JS modules, where many already don't realise it's just another export name string given special syntactical sugar.

guybedford (approved):

README.md
peterhuene: How about `default-export` to also better tie the name to the export list?

peterhuene: Although that might still be confusing. Happy with `default-interface`, either way.

peterhuene: `direct-export`?

guybedford: Maybe just `interface = "foo"`? I assume this feature doesn't compose with `"exports"` anyway?

peterhuene: It's intended to specify which of the exported interfaces gets its functions directly exported by the component itself rather than via an instance export.

`wit-component` currently designates a single interface for exporting directly, the rest go through instance exports, to guarantee there's never a naming conflict and you can easily "cast" a component instance to the appropriate interface at runtime without having to have some special knowledge of how to map exports to interfaces.

guybedford: Just tried it out and it all seems relatively straightforward, thanks for the clarifications. I had thought this option was a direct path itself as opposed to a key name in `"exports"`.

My main concern is just avoiding conflation of default and namespace concepts by keeping terms clear. Of course `default` can mean many things but in this context specifically it would be nice if we can be a bit careful. How about `interface-export`?

peterhuene: A compromise of`direct-interface-export`? We could probably bike shed this all day 😄.

guybedford: The point was very much a comment as opposed to a change request, and it is entirely your decision if you want to keep things as-is or make a change and what name to use. Will mark as resolved for now regardless!

peterhuene: Yeah, I'm definitely convinced that `default` is not conveying enough intent as to be clear, so the feedback is very appreciated.

I didn't want to dictate a name, but let's go with `direct-interface-export` for now and if that proves too verbose or unclear we can update it later.

PR #35, opened 2022-10-10
chore: upgrade to clap v4

ricochet:
Upstream cargo has [upgraded](https://github.com/rust-lang/cargo/blob/master/Cargo.toml#L65) to clap v4.

`takes_value` has been deprecated.
`num_args` (the replacement for `takes_value`) will default `num_args` based on the `ArgAction` and doesn't need to be set for the verbose flag use-case. But there's one more change. Commands with ArgAction::Count must be u8.

Test caught this:

```bash
command=`"/Users/bhayes/repos/bytecodealliance/cargo-component/target/debug/cargo-component" "component" "add"`
code=101
stdout=""
stderr="thread \'main\' panicked at \'assertion failed: `(left == right)`\n  left: `u8`,\n right: `u32`: Argument `verbose`\'s selected action Count contradicts `value_parser` (ValueParser::other(u32))\', /Users/bhayes/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-4.0.12/src/builder/debug_asserts.rs:708:9\nnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace\n"
', /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/ops/function.rs:248:5

---- help stdout ----
thread 'help' panicked at 'Unexpected stdout, failed var.contains(Add a dependency for a WebAssembly component)
├── var:
└── var as str:

command=`"/Users/bhayes/repos/bytecodealliance/cargo-component/target/debug/cargo-component" "component" "help" "add"`
code=101
stdout=""
stderr="thread \'main\' panicked at \'assertion failed: `(left == right)`\n  left: `u8`,\n right: `u32`: Argument `verbose`\'s selected action Count contradicts `value_parser` (ValueParser::other(u32))\', /Users/bhayes/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-4.0.12/src/builder/debug_asserts.rs:708:9\nnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace\n"
', /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/ops/function.rs:248:5
```

`Arg::possible_values` has been replaced with `Arg::value_parser([...])`. Things got a little awkward here. While upstream cargo seems to be able to pass in `value_parser = ["1"]` ([src](https://github.com/rust-lang/cargo/blob/master/src/bin/cargo/commands/metadata.rs#L27)), we get a downcast error.

```bash
stderr="thread \'main\' panicked at \'Mismatch between definition and access of `format_version`. Could not downcast to u32, need to downcast to alloc::string::String\n\', src/commands/metadata.rs:62:25\nnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace\n"
```

Kudos on solid test coverage.
The tests caught all of these after the build succeeded. I added an additional test around format-version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,7 +109,7 @@ dependencies = [
  "bytesize",
  "cargo-platform",
  "cargo-util",
- "clap",
+ "clap 3.2.22",
  "crates-io",
  "crossbeam-utils",
  "curl",
@@ -167,7 +167,7 @@ dependencies = [
  "assert_cmd",
  "cargo",
  "cargo-util",
- "clap",
+ "clap 4.0.12",
  "log",
  "predicates",
  "pretty_env_logger",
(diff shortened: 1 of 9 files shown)


peterhuene (approved):
Thanks for fixing this!

Issue #34, opened 2022-10-10
cargo-component as cargo command for rust-analyzer fails for non-component crates


ricochet:
After setting "cargo-component" as the custom cargo command for rust-analyzer in VS Code, it begins to fail to parse Cargo.toml's for non-cargo component crates.

```bash
"rust-analyzer.server.extraEnv": {
  "CARGO": "cargo-component"
}
```

## Use case 1: workspaces

Cargo.toml of a workspace with only one package ([repo](https://github.com/wasmCloud/wasmcloud-test)):

```toml
[workspace]
members = [ "wasmcloud-test-util"]
```

An example error is:

```bash
Failed to read Cargo metadata from Cargo.toml file /Users/bhayes/repos/wasmCloud/wasmcloud-test/Cargo.toml, None: Failed to run `"cargo-component" "metadata" "--format-version" "1" "--manifest-path" "/Users/bhayes/repos/wasmCloud/wasmcloud-test/Cargo.toml" "--filter-platform" "aarch64-apple-darwin"`: `cargo metadata` exited with an error: error: manifest path `/Users/bhayes/repos/wasmCloud/wasmcloud-test/Cargo.toml` is a virtual manifest, but this command requires running against an actual package in this workspace
```

Running `cargo metadata --format-version 1 --manifest-path /Users/bhayes/repos/wasmCloud/wasmcloud-test/Cargo.toml --filter-platform aarch64-apple-darwin` succeeds. Working directory does not matter.

## Use case 2: package with library

Cargo.toml of a package with a local library ([repo](https://github.com/wasmCloud/wash)):

```bash
Cargo metadata from Cargo.toml file /Users/bhayes/repos/wasmCloud/wash/Cargo.toml, None: Failed to run `"cargo-component" "metadata" "--format-version" "1" "--manifest-path" "/Users/bhayes/repos/wasmCloud/wash/Cargo.toml" "--filter-platform" "aarch64-apple-darwin"`: `cargo metadata` exited with an error: error: manifest `/Users/bhayes/repos/wasmCloud/wash/Cargo.toml` does not contain a `package.metadata` section
```

Running `cargo metadata ...` without "cargo-component" succeeds.

PR #36, opened 2022-10-18
Implement support for workspaces. 

peterhuene:
This PR implements support for handling workspaces and not individual
component packages.

`cargo-component` now scans the set of packages in the workspace for packages
with component metadata, generates bindings for only those packages, and then
also generates components only for those packages (for the build command).

Fixes https://github.com/bytecodealliance/cargo-component/issues/34.

### DIFF
--- a/tests/metadata.rs
+++ b/tests/metadata.rs
@@ -80,7 +80,10 @@ edition = "2021"
     project
         .cargo_component("metadata --format-version 1")
         .assert()
-        .stdout(contains("foo-interface 0.1.0").and(contains("bar-interface 0.1.0")))
+        .stdout(
+            contains("foo-interface 0.1.0")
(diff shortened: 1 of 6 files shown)


tests/metadata.rs
```diff
@@ -42,3 +42,46 @@ fn it_rejects_invalid_format_versions() -> Result<()> {
...
+        .cargo_component("new bar")
+        .assert()
+        .stderr(contains("Created component `bar` package"))
+        .success();
+
+    project
+        .cargo_component("metadata --format-version 1")
+        .assert()
+        .stdout(contains("foo-interface 0.1.0").and(contains("bar-interface 0.1.0")))
+        .success();
```
ricochet, line 84: baz metadata will also need to be printed for rust-analyzer to work. Best to add an assert for that here

ricochet (approved):

PR #37, opened 2022-10-24
Update to latest dependencies.

peterhuene:
This PR updates `cargo-component` to the latest dependencies.

In doing so, `wit-bindgen` has been updated to generate based on a "world" rather than bindings per import/export.

As such, this is a breaking change with user code as now there's a singular "bindings" crate that gets generated.


peterhuene:
Open to anyone that wants, or has the time, to review 😄

guybedford (approved):

PR #41, opened 2022-11-29
Update cargo-component dependencies to latest.

peterhuene:
This PR updates cargo-component's dependencies to latest.

It includes changes to `wit-component` to support the current binary format of the component model.

As a result of changes to wit-parser, interface definitions must now be wrapped in a declaring `interface <name>` scope.

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -1,10 +1,15 @@
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use cargo::{
     ops::{self, NewOptions, VersionControl},
     Config,
 };
 use clap::{ArgAction, Args};
-use std::{fs, path::Path};
+use heck::{ToKebabCase, ToSnakeCase, ToUpperCamelCase};
+use std::{
+    borrow::Cow,
+    fmt, fs,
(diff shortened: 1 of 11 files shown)


src/commands/new.rs
```diff
@@ -1,14 +1,120 @@
...
+    borrow::Cow,
+    fmt, fs,
+    path::{Path, PathBuf},
+};
 use toml_edit::{table, value, Document, InlineTable, Item, Table, Value};
 
 use crate::WIT_BINDGEN_REPO;
 
+fn is_wit_keyword(s: &str) -> bool {
+    // TODO: move this into wit-parser?
```
ricochet, line 18: I think this would be generally useful for other languages so moving to wit-parser makes sense. Shouldn't block getting this change in however

ricochet (approved):

src/commands/new.rs
peterhuene: Agreed.

Ideally the Rust one won't be necessary, either. Currently wit-bindgen-rust doesn't generate raw identifiers for any wit identifiers that are Rust keywords, so I figured we'd preemptively block them here and remove the restriction once it does.

PR #43, opened 2022-12-15
Add a design document around registry integration.

### LINKED ISSUE #42: [Design] Component Registry Dependencies
# Component Registry Dependencies

## Overview

Currently `cargo-component` supports implementing a component by specifying
individual imports and exports defined by local `wit` documents via
`Cargo.toml`.

At the time `cargo-component` was originally implemented, it was imagined that 
a component registry might store individual interface definitions as packages, 
enabling registry dependencies to be specified in a `Cargo.toml` file as:

```toml
[package.metadata.component.imports] 
pkg1 = { version = "1.2.3", package = "org/pkg1" }

[package.metadata.component.exports] 
pkg2 = { version = "1.2.3", package = "org/pkg2" }
```

Therefore at most one interface could be defined in a `wit` document and stored 
in a component registry "interface" package.

As a consequence of this, a [`world`][world] stored in a registry would then 
need to explicitly reference each interface being imported and exported from 
their individual interface packages; this is certainly not ergonomic and would 
definitely contribute to an unnecessary proliferation of packages.

`wit` has since evolved to allow multiple interfaces and worlds to be defined 
in one or more `wit` documents. To facilitate this, 
[`wit` now has syntax for using types][use] from other documents.

With this more flexible approach to defining interfaces, `cargo-component` 
needs a new mechanism for specifying dependencies that are stored in a 
component registry.

This document proposes a design for specifying dependencies from a component 
registry for `cargo-component`.

## Registry package types

Before discussing the proposed design, it might be useful to discuss some 
terminology surrounding the types of packages that might be stored in a 
component registry.

Previously, there was discussion around there being three types of packages in 
a component registry: an _interface_ package, a _world_ package, and a 
_component_ package.

Both _interface_ and _world_ packages are simply WebAssembly components  
containing **only** type information; in terms of the component model proposal, 
the former describes an _instance_ type and the latter describes a _component_ 
type.

A _component_ package is an implementation of a WebAssembly component; thus it 
contains type information _and_ executable code.

With the introduction of the `use` syntax, this proposal suggests reducing the 
types of registry packages to two: a _wit_ package and the 
aforementioned concept of a _component_ package.

A _wit_ package, much like the concepts of _interface_ and _world_
packages, contains only definitions of types, interfaces, and worlds. However, 
a _wit_ package may store any number of definitions, including defining 
both _interfaces_ and _worlds_ in the same package.

## Design overview

_Note: `cargo-component` should support sourcing packages from multiple 
registries, potentially defaulting to a particular registry instance._

This design proposes that `cargo-component` reads only the version 
requirements for component registry dependencies from `Cargo.toml`.

An example `Cargo.toml` of a "greeter" component might look like:

```toml
[package.metadata.component.dependencies]
wasi = "webassembly/wasi:1.2.3"
formatter = "my-org/formatter:3.2.1"
```

The short form of a dependency is `<name> = "<package-id>:<version>"`.

The general form is `<name> = { version = "<version>", package = 
"<package-id>" }`
where `name` here maps to a package name used in `component.wit` (see below) 
and `package-id` is the qualified identifier of the package in a component registry.

Local `wit` documents may still be referenced by using a `path` key instead of the
`package` key.

In this example, `Cargo.toml` is only specifying that version `1.2.3` 
of the `webassembly/wasi` package and version `3.2.1` of the `my-org/formatter` 
package be used.

Note that _what_ is used from the packages is not specified in `Cargo.toml`; 
thus it doesn't describe the _world_ of the component being built in 
any way.

To specify the component's _world_, a `wit` file with a default name of
`component.wit` is used:

```wit
world my-component {
    import wasi-fs: wasi.fs
    import formatter: formatter
    greet: func(name: string)
}
```

In this example, the component imports two interfaces: the `fs` interface from 
a package named `wasi` and the default interface from the `formatter` package. 
The former is used to print the greeting to the console and the latter is used 
to format the message based on whatever formatter implementation
is supplied at runtime.

The component will directly export a function named `greet` that will 
ultimately print a greeting for the given name.

Because `cargo-component` resolves dependencies ahead-of-time, `wit-parser` 
only needs to be instructed where to locate the `wasi` and `formatter` packages to
successfully parse `component.wit`.

## Implementation

To implement this approach, `cargo-component` will parse `Cargo.toml` to figure 
out what component registry dependencies are required.

For consistent builds, it will also consult a lock file (design TBD) for 
specific versions and signatures of the dependencies to use.

`cargo-component` will contact one or more registries to download (or update) 
the package logs of the dependencies, verify the logs, and resolve the version 
requirements to specific versions to download and cache locally.

It will then parse `./component.wit` (the path can be changed in `Cargo.toml`, if 
desired) and provide `wit-parser` with the paths to the cached package 
dependencies.

From this, the definition of a world will be derived  and used to generate the
bindings needed to build the component for commands like
`cargo component build`.

Finally, the resulting component will encode unique `url`s for its imports and
exports based on what packages were resolved (and from where) by
`cargo-component`.

## Benefits

A few benefits to this approach:

  * Dependencies are specified similar to how they are specified for Rust 
    crates in `Cargo.toml`: version requirements go in `Cargo.toml` and 
    what is used from the dependencies is specified in "source code" 
    (`component.wit` in this case).

  * `wit-parser` does not need to be made registry-aware; packages are 
    resolved to local definitions prior to its invocation.

  * Knowledge of a `component.wit` file is transferable to other implementation 
    languages: it is just a `wit` document to describe the component being 
    built and therefore other language-specific tooling would likely also use a 
    `component.wit` file for generating bindings.

## Tooling for other languages

In addition to Rust, this approach can also be adopted for other 
language-specific tooling.

For example, in JavaScript, tooling that wraps `npm` could specify dependency 
version requirements in `package.json` and use a `component.wit` file to 
specify the component type for generating bindings.

Even `wit-bindgen` CLI (or a tool wrapping it) could be made registry-aware by 
sourcing version requirements from a file (`bindgen.toml`?) and use a 
`component.wit` file to generate bindings for supported languages.

[world]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-world
[use]: https://github.com/WebAssembly/component-model/pull/141

peterhuene:
[Rendered view](https://github.com/bytecodealliance/cargo-component/blob/f8bf935bd045b90de6c6482f144217f9b184e114/docs/design/registries.md)

This PR adds a design document for the developer experience surrounding using packages from a component registry in `cargo-component`.

Closes #42.

### DIFF
--- a/docs/design/registries.md
+++ b/docs/design/registries.md
@@ -322,7 +322,7 @@ Because the `targets` specifies a local wit document, the dependencies are
 interpreted as wit external packages only; `regex` and `transcoder` are not 
 imported automatically as in the previous example.
 
-The resulting type for this component will the same as the previous example, 
+The resulting type for this component will be the same as the previous example, 
 but it will also export a `parse` function of type `(string) -> result<_, string>`
 as specified in the world.
 
--- a/docs/design/registries.md
+++ b/docs/design/registries.md
@@ -78,12 +78,8 @@ This design proposes the following tables in `Cargo.toml`:
 
   * `targets` - a target world for the component being authored; this causes 
     bindings to be generated for the specified world's imports and exports.
-    
-  * `world` - a path to a wit document specifying a world definition for the 
-    component being authored; it must contain a `world` with the same name as 
-    the current crate name.
 
-  The `targets` and `world` fields are mutually exclusive.
+  See below for more information on the `targets` field.
 
 * `[package.metadata.component.dependencies]`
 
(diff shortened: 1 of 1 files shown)


docs/design/registries.md
```diff
@@ -0,0 +1,477 @@
...
+  ```toml
+  foo = { package = "ns/foo", version = "0.1.0", registry = "my-registry" }
+  ```
+
+  The `default` name is used when no `registry` field is specified in a
+  dependency. Therefore, specifying a registry with name `default` will 
+  override the built-in default in `cargo-component` (expected to be a 
+  future Bytecode Alliance component registry instance).
+
+  A registry URL may use the `file://` scheme to point at local directory that
```
peterhuene, line 151: We probably want to use some other format for a file-system based registry as the `file` URL scheme can't express immediate relative paths (e.g. `file://foo` isn't technically allowed).

Perhaps instead express a registry entry in the form `name = { path = "<path>" }`.

Thus the form `name = "<registry-url>"` is just a short-hand for `name = { url = "registry-url" }`.

docs/design/registries.md
```diff
@@ -0,0 +1,493 @@
...
+
+```wit
+world my-component {
+  include wasi.command # proposed syntax for including a world in another
+
+  import regex: regex
+  
+  import transcoder: transcoder
+  
+  parse: func(x: string) -> result<_, string>
```
fibonacci1729, line 368:
```suggestion
  export parse: func(x: string) -> result<_, string>
```

docs/design/registries.md
```diff
@@ -0,0 +1,493 @@
...
+http-types = "http/types:1.2.3"
+```
+
+An example `component.wit`:
+
+```wit
+world my-component {
+  import downstream: http-types.handler
+
+  handler: http-types.handler
```
fibonacci1729, line 483:
```suggestion
  export handler: http-types.handler
```

peterhuene:
I have the implementation of this design doc almost ready for PR (plus _a lot_ of new functionality), so I think we can merge this.

docs/design/registries.md
```diff
@@ -0,0 +1,361 @@
...
+  
+  export parse: func(x: string) -> result<_, string>
+}
+```
+
+Because the `targets` specifies a local wit document, the dependencies are 
+interpreted as wit external packages only; `regex` and `transcoder` are not 
+imported automatically as in the previous example.
+
+The resulting type for this component will the same as the previous example,
```
fibonacci1729, line 325:
```suggestion
The resulting type for this component will be the same as the previous example, 
```

fibonacci1729:
One minor nit but otherwise looks great! 💯

PR #46, opened 2023-02-03
Update the registry design doc (again).

peterhuene:
[Rendered version](https://github.com/bytecodealliance/cargo-component/blob/173afdaf83fc1d892215b495a4bd45b8b2fdd60f/docs/design/registries.md)

This PR updates the registry design doc to better describe target worlds.

It now fully separates WIT package dependencies used _only_ for parsing local WIT document targets and the component dependencies.

PR #47, opened 2023-02-04
Initial registry support in `cargo-component`.

peterhuene:
This PR implements a large refactoring of `cargo-component` with the intention of laying the groundwork of supporting component registries.

Additionally, this bumps all crate dependencies to their latest possible versions, including `wasm-tools` and `wit-bindgen`, so that the components produced conform to the latest supported component model binary AST.

The metadata format in `Cargo.toml` has been completely overhauled as a result; older metadata formats will result in errors when parsing the manifest and will require manual conversion to the latest format (sorry!).

Now instead of having individual imported and exported interfaces listed in `Cargo.toml`, a target world is specified that may be a dependency from a registry or a local WIT document (the latter is currently the default of `cargo component new`).

The tool will now write a `Cargo-component.lock` that locks component registry dependencies to specific versions, much like how `Cargo.lock` functions for crate dependencies. 

The `cargo component add` command has been overhauled to work with adding component dependencies from a registry.

A new `cargo component registry` command has been added for creating local file system component registries and publishing packages into local registries.

A `cargo component publish` command, support for remote registries, and a `cargo component update` command will soon be added.


peterhuene:
See #46 for the design document around registry support and the `Cargo.toml` format implemented by this PR.

peterhuene:
Whoops, need to update CI to install a protobuf compiler. I'll fix this tomorrow.

peterhuene:
There was a breaking change to the binary format this week, so we need to update `wit-bindgen` to the latest `wasm-tools` before we can move the deps forward with this PR.

peterhuene:
`wit-bindgen` is preparing for a published release to crates.io, which is currently breaking CI as it impacts the output of `cargo component new`.

I'm going to wait for the published release and include a commit to this PR that both updates `cargo-component`'s dependencies and the output of `cargo component new`.

peterhuene:
@Kylebrown9 thanks for the review! Let's sync up about those proposed changes soon. I have a `cargo component update` command ready to PR as well (thankfully, a *much* smaller change).

PR #50, opened 2023-02-15
Implement a `cargo component update` command.

peterhuene:
This PR implements a `cargo component update` command that will update both the `Cargo.lock` and `Cargo-component.lock` files.


fibonacci1729 (approved):
LGTM

PR #53, opened 2023-02-27
Implement a `--target` option for `cargo component new`.

peterhuene:
This PR implements a `--target` option that causes `cargo component new` to set the target to use in `Cargo.toml`.

It also generates `src/lib.rs` based on the exports of the target world, providing the necessary boilerplate code for users to immediately start implementing the component's exports.

A `--registry` option is also added to set the default registry of the project and is currently used to resolve the `--target` option until an internal default registry is implemented.

Additionally, a `--world` option was added to set the world to use when specifying a target WIT package.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -249,6 +249,7 @@ dependencies = [
  "wat",
  "wit-bindgen-core",
  "wit-bindgen-gen-guest-rust",
+ "wit-bindgen-gen-rust-lib",
  "wit-component 0.7.1",
  "wit-parser",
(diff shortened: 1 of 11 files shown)


peterhuene:
Failure might be some discrepancy with URL parsing on Windows; looking into it.

peterhuene:
Are there any volunteers to take on this PR review? 😄

fibonacci1729 (approved):

peterhuene:
Thanks!

PR #59, opened 2023-03-07
Update installation instructions in the README

### LINKED ISSUE #54: Document system dependencies needed for installation
I ran into an issue when installing that I didn't have protobuf installed on my machine. It would be nice if system dependencies were documented in the README.

I imagine in addition to protobut that a clang/gcc might be necessary (due to the use of the cc crate which might be compiling some c dependencies?). Are there any others? libssl? Happy to send a PR adding this documentation to the README!

peterhuene:
This PR updates the README so that the requirements and installation
sections come first.

Dependencies are also updated to the latest versions.

Closes #54.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,9 +54,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.64"
+version = "0.1.66"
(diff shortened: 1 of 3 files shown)


fibonacci1729 (approved):
LGTM

PR #61, opened 2023-03-14
Move the example world.wit files into wit directories.

sunfishcode:
Add support for specifying a wit directory instead of just a wit file, consistent with what wit-bindgen does [here].

And, move the generated example world.wit file into a wit directory; putting wit files in a wit directory is the [default in wit-bindgen] and used in for example [wasi-http], [wasi-proposal-template], and others, and it keeps the tree tidier when one starts adding more wit files.

[here]: https://github.com/bytecodealliance/wit-bindgen/blob/main/src/bin/wit-bindgen.rs#L147
[wasi-http]: https://github.com/WebAssembly/wasi-http/tree/main/wit
[wasi-proposal-template]: https://github.com/WebAssembly/wasi-proposal-template/tree/main/wit
[default in wit-bindgen]: https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/rust-macro/src/lib.rs#L109


peterhuene (approved):
Looks good!

PR #63, opened 2023-03-14
Add a `doc` subcommand.

sunfishcode:
Add a `cargo component doc` subcommand, which generates API documentation for the user's component and optionally its dependencies.

### DIFF
--- a/src/commands/doc.rs
+++ b/src/commands/doc.rs
@@ -1,0 +1,166 @@
+use crate::{
+    commands::{workspace, CompileOptions, DocOptions},
+    Config,
+};
+use anyhow::Result;
+use cargo::core::compiler::CompileMode;
+use clap::{ArgAction, Args};
+use std::path::PathBuf;
+
+/// Generate API documentation for a WebAssembly component API.
+#[derive(Args)]
+pub struct DocCommand {
+    /// Opens the docs in a browser after the operation
+    #[clap(long = "open")]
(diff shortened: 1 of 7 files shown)


src/commands/doc.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    /// Force generation of all dependency bindings.
+    #[clap(long = "generate")]
+    pub generate: bool,
+}
+
+impl DocCommand {
+    /// Executes the command.
+    pub async fn exec(self, config: &mut Config) -> Result<()> {
+        log::debug!("executing compile command");
```
peterhuene, line 121:
```suggestion
        log::debug!("executing document command");
```

peterhuene (approved):
Awesome, thanks for adding this.

Just the one minor nit.

Additionally, could we add the command to the root `README`?

peterhuene:
Also, could we add a test file for the `doc` command for, at a minimum, testing `--help`, but ideally if we could put a doc comment on the component and check that `target/doc/<name>/index.html` contains its content?

peterhuene (approved):

sunfishcode:
It turns out I added the test wrong, so it's passing for the wrong reason. Fixing this will depend in https://github.com/bytecodealliance/wasm-tools/pull/960 to actually propagate the documentation comment to the bindings, so that it shows up in the generated documentation.

PR #64, opened 2023-03-18
Implement `bin` (command) projects.

peterhuene:
This PR updates `cargo component new` to default to `bin` (command) projects rather than `lib` (reactor) projects, much like cargo does.

This is a breaking change if you've been using `cargo component new` before, as it will now create a "command" that simply has a `main` that prints "hello, world!" instead of a reactor.

To continue to create reactors, pass the `--lib` option to `cargo component new`.

This commit also updates the preview2 adapters for commands/reactors and also all dependencies to latest.


peterhuene:
@sunfishcode since you've been experimenting with targeting WASI with `cargo-component`, would you mind reviewing?

This implements `cargo component new --bin foo` so that `cargo component build` creates an adapted command that prints "hello, world!", instead of a reactor.

src/commands/new.rs
```diff
@@ -272,7 +282,15 @@ impl NewCommand {
         })
     }
 
+    fn is_bin(&self) -> bool {
+        self.bin || !self.lib
+    }
+
     fn generate_source(&self, target: &Option<RegistryPackageResolution>) -> Result<Cow<str>> {
+        if self.is_bin() {
+            return Ok("".into());
```
sunfishcode, line 291: Would it make sense for this string to contain a `fn main() { println!("Hello, World!"); }`?

src/lib.rs
```diff
@@ -198,71 +198,142 @@ fn is_wasm_module(path: impl AsRef<Path>) -> Result<bool> {
...
         .parent()
-        .unwrap()
-        .join("deps")
-        .join(target_path.file_name().unwrap());
+        .expect("expected the output to have a parent directory")
+        .join("deps");
+
+    // Determine what the file would be in the deps directory
+    // We need to do this because cargo will *always* link/copy the dep to
+    // the output directory, even when it's up-to-date.
```
sunfishcode, line 210: I did some quick tests with strace on `cargo-component build` and am not seeing this. I see the `linkat` on the first cargo-component-build run and not subsequent runs when the output dir is up to date. Can you say more about where you're seeing this?

src/commands/new.rs
peterhuene: Currently the underlying `cargo new` populates that for us; this just prevents `cargo component new` from overwriting that.

sunfishcode: Ah, makes sense. Could you add a comment mentioning that?

src/lib.rs
peterhuene: So there are some differences based on platform for what cargo does here.

On Linux, it's probably doing a hardlink and the inodes are the same, so cargo skips making a syscall (there's a test in there if the deps and target paths are the "same" file that bails out successfully).

On macOS, where I am seeing the behavior, it's always doing a copy because of a workaround for an AFS bug. So the inodes differ and cargo always copies the deps over the target even when they're the "same". Because of this, if we don't componentize the deps file, cargo will overwrite the previous component with the module output by the Rust toolchain, which will trigger `cargo component` to componentize it again (and printing a message that it's doing so).

src/commands/new.rs
peterhuene: Will do.

src/lib.rs
peterhuene: I'll at least update the comment to mention that this behavior appears to be macOS-specific.

sunfishcode: Ok. This seems ok for now, though it feels the kind of thing someone may end up revisiting in the future. Users will hit errors because when they end up with multiple versions in their deps directory, and it's possible we may hit other cargo implementation details. So the more we can record about what we know the situation in the comments, the better :-).

peterhuene: Actually, you're right, this isn't a good way of doing this for binaries as cargo will indeed leave the stale ones there after `Cargo.toml` changes, so we shouldn't do it this way.

Let me see if I can figure out exactly what file hash we should be using so we don't search for it.

peterhuene: Alright, there's no good way to do it from cargo's API: the fingerprint calculation is done internal to the compile jobs and doesn't propagate to the compilation result.

I think I'm going to modify this to simply write to the target path, foregoing any deps writing. Doing so also makes it less reliant on cargo internals to make work.

The unfortunate side effect is that, on macOS, users will always see a "creating component" message even if the build was up-to-date.

Perhaps we can work around that by caching mtimes of the files pre-componentization to compare against and forego the message if the mtime matches what was cached previously; we'll still need to componentize since cargo is copying from deps over the previous componentized target, though.

That's a fix for another day.

sunfishcode (approved):

PR #68, opened 2023-03-27
Indicate Required Protoc Min Version

esoterra:
The registry project uses the `optional` keyword and in `proto3` this requires us to use `protoc` version 3.15 or greater where they re-introduced that syntax as per these [release notes](https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0).


peterhuene (approved):

peterhuene:
Thanks!

PR #69, opened 2023-04-06
Implement warg component registry support.

peterhuene:
This PR implements support for warg component registries.

It also removes support for "local" registries and, subsequently, the `registry` subcommand.

The following commands have been implemented:

* `publish` - publishes the current component to a registry.
* `signing new-key` - create a new signing key in the system keyring.
* `signing set-key` - set a signing key in the system keyring.
* `signing delete-key` - delete a signing key from the system keyring.

The tests now require a `warg-server` be on the path to run a local, temporary component registry.


src/registry.rs
```diff
@@ -434,228 +445,574 @@ impl DependencyResolution {
...
-            registry: Box<dyn Registry>,
-            packages: HashMap<&'a PackageId, Vec<PackageEntry<'a>>>,
-        }
+        Ok((self.target_resolutions, self.resolutions))
     }
 
-    /// Iterates over all dependencies of the package.
-    pub fn deps(&self) -> impl Iterator<Item = (&String, &DependencyResolution)> {
-        self.target_dependencies
+    async fn update_and_download(&mut self) -> Result<()> {
```
Kylebrown9, line 867: Separate out update and download into their own functions that pipe data through.

src/commands/signing.rs
Kylebrown9, line 1: Would be more aptly named `key` than `signing`.

Kylebrown9:
Thanks for the PR, it looks great!

src/commands/signing.rs
peterhuene: The new commands are `key new`, `key set`, and `key delete`.

src/registry.rs
peterhuene: I've split this up a little bit better where the update and "download and resolve" steps feed state through.

Let me know if the change doesn't make sense.

Kylebrown9 (approved):
Looks great, thanks!

PR #70, opened 2023-04-11
Fix up-to-date check when target is a wit directory.

### LINKED ISSUE #65: Local target up-to-date check isn't working for directories
If `target` is pointed at a directory (the expectation that it can be parsed as a `wit` package), then this breaks the up-to-date check because it is expecting a path to a file.

As a result, modifying a wit file in the directory doesn't trigger a regeneration of the bindings.

Ideally we would be able to parse the directory from the `wit-parser` API and get back the paths to the files it parsed so that we can include them in the timestamp checks.

peterhuene:
This PR fixes the up-to-date check when the target is a directory containing wit files.

We now parse the wit *before* the up-to-date check, which allows us to determine the additional wit dep paths to timestamp.

Fixes #65.

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -78,6 +78,9 @@ pub struct BindingsGenerator<'a> {
     name: String,
     manifest_path: PathBuf,
     source_path: PathBuf,
+    resolve: Resolve,
+    world: WorldId,
+    deps: Vec<PathBuf>,
 }
 
 impl<'a> BindingsGenerator<'a> {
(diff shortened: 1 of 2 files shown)


peterhuene:
@sunfishcode I know you're in the EU currently, so no rush on a review (also feel free to pass it off); this is a fix for a bug I noticed when we were originally investigating your demo host issue, namely that if you point `cargo-component` at a local directory for the target, changing any of the wits in the directory doesn't trigger a recompilation.

sunfishcode (approved):
Cool.

PR #74, opened 2023-04-27
Update wit-bindgen and wasm-tools dependencies

rylev:
Comes on the heals of https://github.com/bytecodealliance/wit-bindgen/pull/563 and https://github.com/bytecodealliance/wasm-tools/pull/1002

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -456,8 +456,8 @@ dependencies = [
  "warg-client",
  "warg-crypto",
  "warg-protocol",
- "wasm-metadata",
- "wasmparser",
+ "wasm-metadata 0.4.0",
+ "wasmparser 0.103.0",
  "wat",
  "wit-bindgen-core",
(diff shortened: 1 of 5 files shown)


peterhuene:
It looks like this is the first time CI has run since the latest stable Rust release, which changed a compiler diagnostic message the test was expecting; removing `struct` from the message will fix that particular failure.

Looks like `cargo component clippy` is not generating any diagnostics either, so I'm looking into what broke.

peterhuene:
Looks like cargo changed the way it sources environment variables, which the `clippy` subcommand now sets too late in the process to respect.

I'll push the fixes up to this PR.

peterhuene:
Ok fix pushed and I went ahead and updated all of the dependencies in the process.

peterhuene (approved):

peterhuene:
Thanks for updating this, @rylev!

Issue #77, opened 2023-05-10
cargo install fails


schorg:
Wanted to try some component examples, to demonstrate Wasm components,  but

`cargo install --git https://github.com/bytecodealliance/cargo-component`

fails with the following error:

```
error[E0432]: unresolved import `warg_client::storage::PackageStorage`
  --> src/registry.rs:24:44
   |
24 |     storage::{ContentStorage, PackageInfo, PackageStorage},
   |                                            ^^^^^^^^^^^^^^ no `PackageStorage` in `storage`

error[E0599]: no method named `packages` found for reference `&warg_client::Client<FileSystemPackageStorage, FileSystemContentStorage>` in the current scope
   --> src/registry.rs:740:30
    |
740 |                 match client.packages().load_package(id.as_str()).await? {
    |                              ^^^^^^^^ method not found in `&Client<FileSystemPackageStorage, FileSystemContentStorage>`
```

It worked 1 week ago. Do I need to wait until it get's more stable?

schorg:
Compiled it and installed it locally with revision `[eee527f](https://github.com/bytecodealliance/registry/commit/eee527f91d96776b58cf4631911eecabdee7512e)`
of the registry dependency.
and it worked.

Kylebrown9:
Since [`registry #102`](https://github.com/bytecodealliance/registry/pull/102) was merged, the API for accessing registry client storage has changed. Since `cargo-component` uses [git-based dependencies with no revision info](https://github.com/bytecodealliance/cargo-component/blob/main/Cargo.toml#L12) attempting to install `cargo-component` now pulls in an incompatible version of the registry code.

peterhuene:
Also consider using the `--locked` option to ensure that the install follows `Cargo.lock` rather than using the latest dependencies.

We should update the README until such time that install works off solely off of published dependencies.

PR #78, opened 2023-05-11
update to use RegistryStorage from warg

macovedj:
Resolves [this issue](https://github.com/bytecodealliance/cargo-component/issues/77#issuecomment-1542600794)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Build cargo-component
       run: cargo build
     - name: Install warg-server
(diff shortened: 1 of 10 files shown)


peterhuene (approved):

macovedj:
Looking into the failing tests

peterhuene:
Probably need to bump the server revision in the CI yaml.

peterhuene:
FYI, now that warg-server can be used as a library, we should update the tests to run a server in-process and remove this extra dependency on having a warg-server on the PATH (and thus we wouldn't have to update the CI to ensure a matching server implementation).

peterhuene:
@macovedj Thanks for fixing this!

PR #79, opened 2023-05-16
Use wit-bindgen 0.6.0 for generated components

cardoso:
Let me know if I can sync this with the dependency used in the root crate.


peterhuene:
@cardoso thanks for the PR!

Unlike the macros from `wit-bindgen`, `cargo-component` only uses a dependency on `wit-bindgen` in the projects for the runtime functions as code generation is performed in-advance by `cargo-component` itself.

That said, we'll need to update the dependencies for `cargo-component` to match as well.

I have pushed up a commit that does that.

peterhuene (approved):

peterhuene:
Thanks @cardoso!

PR #85, opened 2023-06-06
Implement WIT package dependencies by `path`.

peterhuene:
This PR implements specifying WIT package target dependencies by `path`.

It allows for specifying local target dependencies that are themselves WIT packages, much like how `wit-parser` allows for deps to be found in a `deps` directory.

The `deps` behavior of `wit-parser` can be replicated by explicitly specifying the WIT's top level dependencies in with the following syntax:

```toml
[package.metadata.component.target]
path = "wit"

[package.metadata.component.target.dependencies]
"foo:bar" = { path = "wit/deps/foo-bar" }
"baz:jam" = { path = "wit/deps/baz-jam" }
...
```

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -257,10 +257,9 @@ publish = false
     fn create_target_world(
         resolution: &PackageDependencyResolution,
     ) -> Result<(Resolve, WorldId, Vec<PathBuf>)> {
-        let (mut merged, world_id, deps) = match &resolution.metadata.section.target {
+        let (mut merged, world_id, mut deps) = match &resolution.metadata.section.target {
             Some(Target::Package { id, world, .. }) => {
-                let (merged, world) = Self::target_package(resolution, id, world.as_deref())?;
-                (merged, world, Vec::new())
+                Self::target_package(resolution, id, world.as_deref())?
             }
             Some(Target::Local { path, world, .. }) => {
                 Self::target_local_path(resolution, path, world.as_deref())?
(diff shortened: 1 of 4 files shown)


fibonacci1729 (approved):

fibonacci1729:
One possible future gotcha here is that a target dependency *must* be a directory to use WIT; i.e. single file WIT packages won't work since they are assumed to be encoded. Maybe a note in the documentation somewhere could help to track this?

sunfishcode:
Would it make sense to make this behave like: if a directory named `deps` exists, call `Resolve::push_dir` for it, otherwise don't and use the registry? And maybe also have it print an informational message like "Using the local `deps` directory instead of the registry" or so? That would reduce boilerplate for some common use cases.

peterhuene:
> One possible future gotcha here is that a target dependency must be a directory to use WIT; i.e. single file WIT packages won't work since they are assumed to be encoded. Maybe a note in the documentation somewhere could help to track this?

We could certainly peak at the file to see if it is binary encoded, otherwise parse as a single file wit package; that'd certainly make it more consistent.

> Would it make sense to make this behave like: if a directory named deps exists, call Resolve::push_dir for it, otherwise don't and use the registry? And maybe also have it print an informational message like "Using the local deps directory instead of the registry" or so? That would reduce boilerplate for some common use cases.

We could go that route where if the target is a path (if it's a file, use the file's directory) and it contains a `deps` directory, we just call `Resolve::push_dir` and call it a day, perhaps emit a warning if the target dependencies list is non-empty as they won't be respected.

It can be said that the explicitness does follow more of the "cargo way" of specifying dependencies rather than scanning the file system, but I don't have strong opinions here.

peterhuene:
@fibonacci1729 I pushed up a commit to allow for single WIT files to be parsed as a target dependency specified by `path`.

fibonacci1729:
Looks good to me!

sunfishcode:
Me too! That being explicit is consistent with cargo overall is a good point. I'm content to wait and see how this feels in practice.

sunfishcode (approved):

Issue #89, opened 2023-06-09
Code generated by the `--target` option of the `new` command should have use declarations


peterhuene:
Currently, the source code generated by the `--target` option of the `new` command fully qualifies every type name when generating the trait implementations of any exports for the target world.

Instead, it should be smarter and generate `use` declarations of the referenced types, falling back to a qualified path only when there would be a conflict.

peterhuene:
I have this implemented, but need #88 merged first.

peterhuene:
Example output from a managed `wasi:http/proxy` target (introduced the `custom` and `custom2` exports to show a conflicting name):

```rust
use bindings::exports::wasi::http::{
    custom::{Custom, InputStream, OutputStream},
    custom2::{Custom2, Pollable},
    incoming_handler::{IncomingHandler, IncomingRequest, ResponseOutparam},
};

struct Component;

impl IncomingHandler for Component {
    fn handle(request: IncomingRequest, response_out: ResponseOutparam) {
        unimplemented!()
    }
}

impl Custom for Component {
    fn foo(input: InputStream) -> OutputStream {
        unimplemented!()
    }
}

impl Custom2 for Component {
    fn foo(p: Pollable) -> bindings::exports::wasi::http::custom2::OutputStream {
        unimplemented!()
    }
}

bindings::export!(Component);
```

PR #90, opened 2023-06-10
Implement use declarations for `--target` option.

peterhuene:
This PR implements generating `use` declarations for the `--target` option
of the `new` command.

Previously, all types from the bindings were using fully-qualified paths.

This implementation uses a simplistic trie to generate `use` declarations of
types from the bindings.

Conflicting type names result in fully qualified paths for any types that
conflict with existing `use` declarations.

Also bumps the `wit-bindgen` version output from the `new` command to 0.7.0.

Closes https://github.com/bytecodealliance/cargo-component/issues/89.


src/bindings.rs
```diff
@@ -643,32 +644,173 @@ publish = false
...
+            let (index, _) = self.segments.get_full(segment)?;
+            node = node.children.get(&index)?;
+        }
+
+        Some(node.tys.iter().map(|i| self.types[*i].as_str()))
+    }
+
+    /// Inserts a new use of the given type.
+    ///
+    /// This method handles the proper casing for segment paths and type names.
```
peterhuene, line 711:
```suggestion
    /// This method handles the proper casing for path segments and type names.
```

src/bindings.rs
```diff
@@ -643,32 +644,173 @@ publish = false
...
-                ns = pkg.name.namespace.to_snake_case(),
-                pkg = pkg.name.name.to_snake_case(),
-                name = name.to_snake_case(),
+/// Represents a node in a "use" trie.
+#[derive(Default)]
+struct UseTrieNode {
+    // Map of segment index to trie node
+    children: HashMap<usize, UseTrieNode>,
+    // Set of type indexes that are used at this node
+    tys: HashSet<usize>,
```
sunfishcode, line 653: Since the iteration order of `tys` determines what order the types are printed in, could these use `indexmap::IndexMap` and `indexmap::IndexSet` here instead, so that the output is deterministic?

src/bindings.rs
```diff
@@ -643,32 +644,173 @@ publish = false
...
+/// A trie that maintains uses of types when generating Rust source code.
+#[derive(Default)]
+struct UseTrie {
+    /// The root trie node.
+    root: UseTrieNode,
+    /// The set of path segments.
+    segments: IndexSet<String>,
+    /// The set of type names.
+    types: IndexSet<String>,
+}
```
sunfishcode, line 695: Would it be feasible to move `UseTrie` and `UseTrieNode` into a separate source file? That'd keep src/bindings.rs a little tidier for people studying how bindings generation works at a high level.

sunfishcode (commented):
`rustfmt` has [an option](https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#imports_granularity) to perform this trie factoring automatically; `--config imports_granularity=Module`. Since cargo component already runs rustfmt (by default), would it make sense to enable that option instead of using a trie?

peterhuene:
@sunfishcode although `rustfmt` does support organizing the `use` directives (it'd be similar to the `Crate`  option, I believe), but I didn't want to rely on it as may not be installed or on the PATH when running the command (it's optional).

src/bindings.rs
peterhuene: Originally I left the order unstable as `rustfmt` would sort them, but this should be fixed if we don't want to rely on `rustfmt` with similar justification as to why use a trie in the first place.

I'll fix this.

fibonacci1729 (approved):
LGTM! I agree with teasing out the `UseTrie` and friends into a separate module and not taking a hard dependency on `rustfmt`.

sunfishcode (approved):
Makes sense!

PR #92, opened 2023-06-21
Export wasi adapter version even if git fails

figsoda:
The package fails to build when it is not in a git repository due to the usage of `env!("WASI_ADAPTER_VERSION")`, this fixes the issue


peterhuene (approved):
Thanks for fixing this!

PR #95, opened 2023-07-04
Update README.md to clarify installation requirements

mikkelhegn:


peterhuene:
Hi @mikkelhegn and thanks for the PR!

I'm going to close this in favor of #96 as `protoc` is going to be removed as a install/build dependency for `cargo-component` as a result.

Thanks!

PR #99, opened 2023-07-20
Fix structure of `Resolve` with an empty world

alexcrichton:
This fixes a mistake where the `Resolve` synthesized for an empty WIT package had a world but the world wasn't listed in the `Package` which caused an invalid custom section to be produced where no world was emitted. The world injected is registered with its parent package here to establish the relationship which avoids a bad custom section.


peterhuene (approved):
Thanks!

PR #100, opened 2023-07-21
Add an optional configuration to disable `rustfmt`

alexcrichton:
Currently `wit-bindgen` is a relatively dumb code generator and the generated code can fail in `rustfmt` due to long lines and things like that. I'm not sure how best to fix this in `wit-bindgen` or how best to communicate this to `rustfmt` (or if it's even a bug there). One option would be to ignore `rustfmt` failures in `wit-bindgen` but that seems not great because then there's not a great means to communicate the non-fatal error.

In lieu of solving some of these thornier issues I figured it'd be good to at least have an escape hatch in `cargo component` to disable `rustfmt` in case errors crop up again in the future.

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -222,7 +222,12 @@ publish = false
         })?;
 
         let opts = Opts {
-            rustfmt: true,
+            rustfmt: self
+                .resolution
+                .metadata
+                .section
+                .rustfmt_bindings
+                .unwrap_or(true),
             macro_export: true,
             macro_call_prefix: Some("bindings::".to_string()),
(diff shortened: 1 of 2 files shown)


peterhuene:
I have a (huge) refactoring to cargo-component incoming that will eliminate the need for this PR as we won't be generating bindings into a separate crate (among other big changes).

I'm fine with merging this now if it's blocking anyone, but I'm hoping that by early next week it'll be entirely unnecessary.

alexcrichton:
Nah no worries that sounds good! This isn't urgent so I think it's definitely ok to wait for that 👍

PR #101, opened 2023-07-25
Refactor `cargo-component` internals and introduce proc-macro.

### LINKED ISSUE #91: Support expanding bindings macro
I'd like to be able to expand the bindings macro as well as any other macros I might be using to debug. I frequently use [cargo-expand](https://github.com/dtolnay/cargo-expand) for this.

### LINKED ISSUE #97: Notice: upcoming breaking change
## What

A breaking change to how `cargo-component` generates bindings will be landing soon.

Manual intervention will be required to keep existing component projects building with a newer `cargo-component`.

When #101 lands, do the following to update your component projects:

1) Remove the dependency on `wit-bindgen`: `cargo rm wit-bindgen`.
2) Add a dependency on `cargo-component-bindings`: `cargo add cargo-component-bindings`.
3) Add a `cargo_component_bindings::generate!();` to the top of your component's source file.
4) Remove any `export!` macro invocations from your source.

## Why

An upcoming change to `wit-bindgen` (see bytecodealliance/wit-bindgen#604) to support resources from the component model will preclude the manner in which `cargo-component` generates its bindings today.

To fully support resources, `wit-bindgen` will need to be provided the Rust types implementing the resources at the time of bindings generation; additionally, the generated `export!` macro is being removed in favor of also supplying the implementing Rust type at generation time.

Ultimately this means that `cargo-component` can no longer generate bindings into a separate crate and automatically inject a dependency on the bindings crate from the component crate.

However, the changes to `cargo-component` are not dependent on the support for resources in `wit-bindgen` and will likely land ahead of resources support.

## How

Instead of generating bindings into a `bindings` crate, `cargo-component` will adopt a proc-macro approach similar to `wit-bindgen::generate!`.

The `cargo-component` version of this macro will differ from the `wit-bindgen` macro in that it will generate bindings from an encoding of the target world resulting from dependency resolution rather than parsing a local wit package.

Currently, a generated `bindings` crate is used for the bindings; with these changes, bindings will be generated into a `bindings` _module_ in the current crate.

Here is an example of defining a component in Rust with `cargo-component` today:

```rust
use bindings::Example;

struct Component;

impl Example for Component {
    fn hello_world() -> String {
        "Hello, World!".to_string()
    }
}

bindings::export!(Component);
```

And here is what it will look like with the upcoming changes:

```rust
cargo_component_bindings::generate!();

use bindings::Example;

struct Component;

impl Example for Component {
    fn hello_world() -> String {
        "Hello, World!".to_string()
    }
}
```

A positive side effect of these changes is that it should improve the user experience if the project is built for the first time with `cargo build` instead of `cargo component build`: a compiler error will be emitted instructing users to build with `cargo component build` and a link to where it can be installed.

Additionally, as a magic crate dependency is no longer added by `cargo-component`, tools like `rust-analyzer` and `cargo expand` (see #91) will now work provided that their invocation follows a `cargo-component` command that resolves dependencies to generate bindings information.

## Where

Right here, silly.

## When

soon

peterhuene:
This PR includes a *very* large refactoring to the internals of `cargo-component` and how it works.

The biggest change is that `cargo-component` no longer generates an external bindings crate that is automatically added as a dependency to a user's component project. The motivation behind this change is that support for resources from the component model will preclude the ability to reference types from the user's crate because the `export!` macro is being removed.

Instead, component projects will need to reference the `cargo-component-bindings` crate to invoke the `generate!` proc-macro. This macro will expand to the bindings code based on the target world encoded from dependency resolution performed by `cargo-component`.

By default, the `generate!` macro will assume that the world / all exported interfaces are implemented by a type named `Component`, however that can be changed with an argument to the macro.

With the removal of the need for an external bindings crate, `cargo-component` no longer has a concrete reason to directly depend on the `cargo` crate as a dependency. Instead, it now spawns `cargo` for all subcommands that aren't built-in to `cargo-component`.

For this to function, `cargo-component` detects certain `cargo` options that may alter output (`--verbose`, `--quiet`, etc.) or what gets built (`--manifest-path`, `--target`, etc.). Before any unrecognized commands are passed to `cargo`, `cargo-component` will update the encoding of target worlds for bindings allowing commands that analyze or build the source code to work.

Additionally, `cargo-component` will recognize certain `cargo` subcommands that have outputs that need to be componentized.

As a result of these changes, about 150 crate dependencies were removed, the `cargo-component` executable shrunk about 12 MiB in size, and commands such as `cargo component expand` now "just work" without `cargo-component` having to provide its own implementation.

All dependencies, and the WASI adapters, are updated as part of these changes.

Note: the `test` command still isn't implemented.

Closes #97.
Closes #91.


peterhuene:
Additionally, this PR also refactors some of the `cargo-component` internals into a "core" crate.

This will allow for sharing implementation with `cargo-component` and the `wit` tool from https://github.com/peterhuene/wit (which I plan to merge into this repo shortly).

README.md
```diff
@@ -159,27 +153,27 @@ and tooling developers can use to test their implementations.
...
+When the Rust compiler supports a [_preview2_ version of the WASI target][1],
 support in `cargo component` for adapting a _preview1_ module will be removed.
 
 [1]: https://github.com/rust-lang/compiler-team/issues/594
 [2]: https://github.com/bytecodealliance/preview2-prototyping/tree/main/wasi-common
 
 ## Getting Started
 
-Use `cargo component new --lib <name>` to create a new reactor component.
+Use `cargo component new --reactor <name>` to create a new reactor component.
```
sunfishcode, line 176: If we're going to say "reactor" here, it'd be good to have a brief explanation of what that means. Perhaps, "... which is a component that doesn't have a `run` function, and is meant to be used as a library".

README.md
```diff
@@ -190,95 +184,85 @@ package my-org:my-component
...
 ```
 
-Here `bindings` is the bindings crate that `cargo component` generated for you.
+The `generate!` macro is responsible for generating the bindings to allow the
+Rust code to export what is expected of the component.
 
-The `export!` macro informs the bindings that the `Component` type exports
-all interfaces listed in `Cargo.toml`.
+It generates a Rust module named `bindings` containing the types and traits the
+correspond to the world definition.
```
sunfishcode, line 213: One slightly subtle consequence of the change from a bindings crate to a bindings module is that the generated module is not accessible outside the module that invokes the macro. That is, you get a `mod`, and sometimes a `pub(crate) mod` would be useful. I'm not sure offhand what to suggest for that though.

README.md
```diff
@@ -190,95 +184,85 @@ package my-org:my-component
...
 * `cargo component publish` - publishes a WebAssembly component to a [warg](https://warg.io/)
   component registry.
-* `cargo component key` - manages signing keys for publishing WebAssembly 
+* `cargo component key` - manages signing keys for publishing WebAssembly
   components.
-* `cargo component wit` - manages the target WIT package of the component.
 
-More commands will be added over time.
+Unrecognized commands are passed through to `cargo` itself, but only after the 
+bindings information for component packages has been updated.
```
sunfishcode, line 230: I myself would find it useful to if this mentioned "build". "check", "doc", and "clippy" as examples of commands that are passed through.

crates/core/src/progress.rs
```diff
@@ -0,0 +1,524 @@
...
+}
+
+/// A progress bar implementation.
+pub struct ProgressBar<'a> {
+    state: Option<State<'a>>,
+}
+
+/// Indicates the style of information for displaying the amount of progress.
+///
+/// See also [`Progress::print_now`] for displaying progress without a bar.
```
sunfishcode, line 25:
```suggestion
/// See also [`ProgressBar::print_now`] for displaying progress without a bar.
```

crates/core/src/progress.rs
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    /// Returns whether or not the progress bar is allowed to be displayed.
+    pub fn is_enabled(&self) -> bool {
+        self.state.is_some()
+    }
+
+    /// Creates a new `Progress` with the [`ProgressStyle::Percentage`] style.
+    ///
+    /// See [`Progress::with_style`] for more information.
```
sunfishcode, line 129:
```suggestion
    /// See [`ProgressBar::with_style`] for more information.
```

crates/core/src/progress.rs
```diff
@@ -0,0 +1,524 @@
...
+        if !s.throttle.allowed() {
+            return Ok(());
+        }
+
+        s.tick(cur, max, msg)
+    }
+
+    /// Updates the state of the progress bar.
+    ///
+    /// This is the same as [`Progress::tick`], but ignores rate throttling
```
sunfishcode, line 171:
```suggestion
    /// This is the same as [`ProgressBar::tick`], but ignores rate throttling
```

crates/core/src/progress.rs
```diff
@@ -0,0 +1,524 @@
...
+        match self.state {
+            Some(ref mut s) => s.tick(cur, max, msg),
+            None => Ok(()),
+        }
+    }
+
+    /// Returns whether or not updates are currently being throttled.
+    ///
+    /// This can be useful if computing the values for calling the
+    /// [`Progress::tick`] function may require some expensive work.
```
sunfishcode, line 187:
```suggestion
    /// [`ProgressBar::tick`] function may require some expensive work.
```

sunfishcode (commented):
I'm not very familiar with the registry parts, but overall this looks good! My main concern is just how we can make the bindings module accessible to other modules:

README.md
peterhuene: I think a `pub(crate)` on the bindings module would make sense and if they need to be public for some reason, the module containing the `generate!` macro can reexport them.

sunfishcode (approved):

peterhuene:
@sunfishcode thanks for looking this huge change over!

PR #103, opened 2023-07-28
Implement a WIT package tool.

peterhuene:
This PR implements a WIT package tool called `wit`.

The tool may be used to initialize, build, and publish a WIT package to a registry.

The tool shares the UI and dependency resolution code from `cargo-component`.


peterhuene:
I'll push up some tests for the new tool shortly.

peterhuene:
Just a few incorrect test assertions on Windows due to a different path separator (sigh) to fix.

crates/core/src/keyring.rs
```diff
@@ -1,67 +1,65 @@
...
-    .context("failed to get keyring entry")
+pub fn get_signing_key_entry(registry_url: &RegistryUrl, key_name: &str) -> Result<Entry> {
+    let label = format!("warg-signing-key:{}", registry_url.safe_label());
+    Entry::new(&label, key_name).context("failed to get keyring entry")
 }
 
-/// Gets the signing key for the given registry host and key name.
-pub fn get_signing_key(host: &str, name: &str) -> Result<PrivateKey> {
-    let entry = get_signing_key_entry(host, name)?;
+/// Gets the signing key for the given registry registry_label and key name.
```
sunfishcode, line 15: Minor nit but it's a littie inconsistent that registry_label is spelled with an underscore and "key name" is not here.

crates/wit/README.md
```diff
@@ -0,0 +1,204 @@
...
+`wit` is a (future) [Bytecode Alliance](https://bytecodealliance.org/)
+project, and follows the Bytecode Alliance's [Code of Conduct](CODE_OF_CONDUCT.md)
+and [Organizational Code of Conduct](ORG_CODE_OF_CONDUCT.md).
+
+### Getting the Code
+
+You'll clone the code via `git`:
+
+```
+git clone https://github.com/peterhuene/wit
```
sunfishcode, line 165: Should this be updated to the wit directory in the bytecodealliance repo?

crates/wit/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+[registries]
+```
+
+By default, the WIT package will not have any dependencies specified.
+
+The registries section contains a mapping of registry names to URLs.
+
+A registry named `default` will be the registry to use when a dependency does
+not explicitly specify a registry.
```
sunfishcode, line 43: It may be worth mentioning that the reason we need this is that the design is for federated registries, rather than having a single central registry that gets used by default.

crates/wit/README.md
```diff
@@ -0,0 +1,204 @@
+# The `wit` tool
+
+A tool for creating and publishing WIT packages to a [WebAssembly component
+registry](https://warg.io/).
+
+WIT packages are used in the [WebAssembly Component Model](https://github.com/WebAssembly/component-model/)
+for defining interfaces and types used in WebAssembly components.
```
sunfishcode, line 7: In addition to interfaces and types, WIT files also defines worlds.

crates/wit/README.md
```diff
@@ -0,0 +1,204 @@
...
+version = "0.1.0"
+
+[dependencies]
+
+[registries]
+```
+
+By default, the WIT package will not have any dependencies specified.
+
+The registries section contains a mapping of registry names to URLs.
```
sunfishcode, line 40: It feels like this would benefit from an example somewhere in this doc of what this syntax looks like.

sunfishcode (commented):
This looks great!

sunfishcode (approved):

peterhuene:
Thanks for the review @sunfishcode!

PR #107, opened 2023-08-03
Implement support for resources.

peterhuene:
This PR implements support for resource types in the `generate!` macro.

Note this is dependent upon bytecodealliance/wit-bindgen#635 to be fixed before the new tests will pass.

Additionally, we'll gate merging this on a new release of wit-bindgen so that we don't have to reference the dependencies by git.


peterhuene:
This should now pass CI and be ready for review.

sunfishcode (commented):
Could you add a test that uses the `ownership:` option? Otherwise this looks good to me!

sunfishcode (approved):

PR #108, opened 2023-08-15
Fix update commands.

peterhuene:
This PR fixes both `cargo component update` and `wit update` to properly print out added/removed messages for dependencies that might have changed in `Cargo.toml`/`wit.toml`, rather than panic.

### DIFF
--- a/crates/wit/src/lib.rs
+++ b/crates/wit/src/lib.rs
@@ -416,23 +416,53 @@ pub async fn update_lockfile(
 
     let new_lock_file = to_lock_file(&map);
 
-    // The lock file doesn't have transitive dependencies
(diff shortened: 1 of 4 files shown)


elliottt (approved):
This looks great to me!

You mentioned in a DM refactoring to share the code for reporting what's added/removed from the lock files, and I think that would be my only suggestion here :)

peterhuene:
Indeed, I plan on doing that refactoring soon, but there are subtle differences in the two code based (mostly around `cargo-component` having `--locked`, `--offline`, and `--frozen` options which the `wit` tool does not; the refactoring would probably add those options where applicable).

PR #109, opened 2023-08-16
Add the `--target` flag to `cargo component add`

elliottt:
Adds the `--target` flag to `cargo component add`, allowing the `package.metadata.component.target.dependencies` section to be manipulated by the `add` command.

### DIFF
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -1,5 +1,7 @@
 use crate::{
-    config::CargoPackageSpec, load_component_metadata, load_metadata, metadata::ComponentMetadata,
+    config::CargoPackageSpec,
+    load_component_metadata, load_metadata,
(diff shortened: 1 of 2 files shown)


src/commands/add.rs
```diff
@@ -46,6 +48,10 @@ pub struct AddCommand {
     /// The id of the package to add a dependency to.
     #[clap(value_name = "PACKAGE")]
     pub package: VersionedPackageId,
+
+    /// Add dependencies to the target component
```
peterhuene, line 52:
```suggestion
    /// Add the dependency to the list of target dependencies
```

peterhuene (commented):
Just the one nit for the CLI help, otherwise looks great 👍

peterhuene (approved):

PR #112, opened 2023-08-16
Refactor the commands to prevent cargo option parsing for built-ins.

peterhuene:
This PR refactors the commands such that the built-in commands do not parse cargo options.

It will allow the built-in commands to fully define what options they support irrespective of what options cargo might support (e.g. `target`).


elliottt (approved):
This looks great, thank you for the quick fix!

elliottt (approved):

PR #113, opened 2023-08-16
Add `--path` to `cargo component add`

elliottt:
Add the `--path` flag to `cargo component add`, allowing dependencies to be resolved from local files. In this case, the version is defaulted to `*`.

### DIFF
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -13,7 +13,10 @@ use cargo_component_core::{
 use cargo_metadata::Package;
 use clap::Args;
 use semver::VersionReq;
-use std::{fs, path::PathBuf};
+use std::{
+    fs,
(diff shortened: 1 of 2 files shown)


src/commands/add.rs
```diff
@@ -52,6 +55,10 @@ pub struct AddCommand {
     /// Add the dependency to the list of target dependencies
     #[clap(long = "target")]
     pub target: bool,
+
+    /// Add a package dependency to this directory.
```
peterhuene, line 59:
```suggestion
    /// Add a package dependency to a file or directory.
```

peterhuene (commented):
Just the one minor nit for the CLI help text, as it also supports pointing at a single WIT or WASM file.

peterhuene (approved):

PR #114, opened 2023-08-18
Update deps to latest.

peterhuene:


peterhuene:
Sigh, looks like we'll have to insert the patch into every generated test project. Let's see if we can do a release of wit-bindgen/wasm-tools instead.

peterhuene:
@elliottt can I trouble you for a quick :+1: to unblock the other PRs? Thanks!

elliottt (approved):
:+1:

PR #115, opened 2023-08-18
Implement generating target code for resources.

peterhuene:
This PR implements the missing support in generating code for the `--target` option of `cargo component new` when the targeted world exports items with owned and borrowed resources.


peterhuene:
Expecting CI to fail until the wit-bindgen/wasm-tools deps are sorted out.

elliottt (approved):

PR #116, opened 2023-08-18
Add `--path` option to `wit add` command.

peterhuene:
This PR adds a `--path` option to the `wit add` command, similar to how PR #113 added the option to `cargo component add`.

It also fixes the formatting such that dependencies that need an inline table are written with an inline table.


peterhuene:
Ignore the CI failures for now, once wasm-tools and wit-bindgen are released, I'll update deps and rebase the outstanding PRs.

elliottt (approved):
This looks great!

PR #130, opened 2023-08-30
Remove implicit versioning in WIT resolves.

### LINKED ISSUE #111: `cargo-component` overwrites WIT package version with `Cargo.toml` version
`cargo-component` seems to automagically set a component's interface version to be that of the Cargo project rather than using the version (or lack of version specified in the WIT interface). 

## What works
The standard "hello, world" example that does not use an interface works as expected:

```bash
$ cargo component new --reactor example
$ cd example
$ cargo component build
$ wasm-tools print target/wasm32-wasi/debug/example.wasm | tail
(type (;13;) (func (result string)))
  (alias core export 2 "hello-world" (core func (;33;)))
  (alias core export 2 "cabi_post_hello-world" (core func (;34;)))
  (func (;14;) (type 13) (canon lift (core func 33) (memory 0) string-encoding=utf8 (post-return 34)))
  (export (;15;) "hello-world" (func 14))
  (@producers
    (processed-by "wit-component" "0.13.1")
    (processed-by "cargo-component" "0.1.0 (26e55d8 2023-08-15 wasi:2750b73)")
  )
)%

```
Function `hello-world` is exported as expected. 

## What does not work
However, if the WIT is updated to export the `hello-world` function from an interface, `cargo-component` automatically appends a version suffix to the interface to match that of the `Cargo.toml`. For example, if you change the WIT to:

```bash
package component:example

interface foo {
    hello-world: func() -> string
}

world example {
    export foo
}
```

And update the implementation in `lib.rs` to:

```bash
cargo_component_bindings::generate!();

use bindings::exports::component::example::foo::Foo;

struct Component;

impl Foo for Component {
    /// Say hello!
    fn hello_world() -> String {
        "Hello, World!".to_string()
    }
}
```

After rebuilding, you will see that the version `0.1.0` is attached to the interface automagically by `cargo-component`:

```bash
$ cargo component build
$ wasm-tools print target/wasm32-wasi/debug/example.wasm | tail
(instance (;8;) (instantiate 0
      (with "import-func-hello-world" (func 14))
    )
  )
  (export (;9;) (interface "component:example/foo@0.1.0") (instance 8))
  (@producers
    (processed-by "wit-component" "0.13.1")
    (processed-by "cargo-component" "0.1.0 (26e55d8 2023-08-15 wasi:2750b73)")
  )
)%
```

After updating the version in `Cargo.toml` to `"0.2.0"` and rebuilding, the version is now `0.2.0`:

```bash
$ cargo component build
$ wasm-tools print target/wasm32-wasi/debug/example.wasm | tail
(instance (;8;) (instantiate 0
      (with "import-func-hello-world" (func 14))
    )
  )
  (export (;9;) (interface "component:example/foo@0.2.0") (instance 8))
  (@producers
    (processed-by "wit-component" "0.13.1")
    (processed-by "cargo-component" "0.1.0 (26e55d8 2023-08-15 wasi:2750b73)")
  )
)%
```

Even if i set the package version in the WIT `package component:example@0.5.0`, the version from the `Cargo.toml` is set on the component interface.

This becomes problematic when writing a Rust program that uses the component, referencing the WIT (`wasmtime::component::bindgen!("example" in "world.wit");`). It cannot find the instance due to mismatched versioning and errors with:

```bash
Error: instance 'example:service/handler' not found
```

### LINKED ISSUE #129: Versioning in bindings
Currently, `cargo-component` injects the component version (sourced from `Cargo.toml`) into a locally defined target world when the `package` directive does not contain a version (see #111).

The intention behind this is so that `Cargo.toml` serves as the source of truth for the "version" of the component and, in the case of targeting a local world definition, its interfaces as well.

However, as documented in #111, this runs afoul of Wasmtime's `bindgen!` macro because it only supports reading local WIT files to generate the host bindings and only the `package` directive conveys the version of the interfaces.

The result is a mismatch between the two where the version implicitly supplied by `cargo-component` is not the same as the version (usually missing) used in the host bindings and the host can't load the component.

A larger question should be asked of the component model proposal: where should version information actually go in interfaces? Right now it participates as part of the extern name, which means that the version must be an _identical_ string match even when the instance might be _type compatible_. This seems incorrect on the surface.

For `cargo-component`, the question is: how do we make this a good experience while still maintaining the goals behind versioning artifacts it produces?

One option is to stop automatically injecting the component version into the target world and let the two have independent versions so that what `cargo-component` and Wasmtime's `bindgen!` generate for bindings match; I'm in favor of doing this.

Another problem is what to do about target dependencies referenced from a registry?

For WasmCon, the hope is to publish WASI packages to a registry which can be easily targeted with `cargo component new --target` or used in a local world definition with `cargo component add --target`.

As those dependencies are definitively versioned (they must be to exist in the registry), the bindings `cargo-component` generates will also be versioned.

But the WIT files that were used to create the adapter modules and WASI support within Wasmtime are _not_ versioned to match, so would not be compatible based solely on _extern name_.

I propose that `cargo-component` stops overriding the version specified in the local target definition based on the version resolved from the registry; if the version is specified in WIT (and it matches the resolved version), then it is used; otherwise if no version was specified, then there's no version used in the bindings either.

peterhuene:
This PR changes `cargo-component` so that it no longer implicitly sets the version numbers on any local WIT definitions.

Closes #129.
Closes #111.

### DIFF
--- a/crates/wit/src/lib.rs
+++ b/crates/wit/src/lib.rs
@@ -12,7 +12,6 @@ use cargo_component_core::{
 use config::Config;
 use indexmap::{IndexMap, IndexSet};
 use lock::{acquire_lock_file_ro, acquire_lock_file_rw, to_lock_file};
-use semver::Version;
 use std::{
     collections::{HashMap, HashSet},
     path::Path,
@@ -116,7 +115,7 @@ fn parse_wit_package(
     }
 
     // Parse the root package itself
-    let mut root = UnresolvedPackage::parse_dir(dir).with_context(|| {
+    let root = UnresolvedPackage::parse_dir(dir).with_context(|| {
         format!(
             "failed to parse package from directory `{dir}`",
             dir = dir.display()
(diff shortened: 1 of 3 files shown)


Kylebrown9 (approved):
LGTM! I think this will be a less surprising behavior for most users.

PR #132, opened 2023-09-01
Add registry metadata to published components.

peterhuene:
This PR adds registry metadata to components published to a registry with
`cargo component publish` and `wit publish`.

### DIFF
--- a/crates/wit/src/commands/add.rs
+++ b/crates/wit/src/commands/add.rs
@@ -1,9 +1,4 @@
-use std::path::PathBuf;
-
-use crate::{
-    config::{Config, CONFIG_FILE_NAME},
-    resolve_dependencies,
-};
+use crate::config::{Config, CONFIG_FILE_NAME};
 use anyhow::{bail, Context, Result};
 use cargo_component_core::{
     command::CommonOptions,
@@ -13,6 +8,7 @@ use cargo_component_core::{
 };
 use clap::Args;
 use semver::VersionReq;
+use std::path::PathBuf;
 use warg_protocol::registry::PackageId;
 
 async fn resolve_version(
@@ -126,16 +122,6 @@ impl AddCommand {
                 config
                     .dependencies
                     .insert(id.clone(), Dependency::Package(package));
-
-                // Resolve all dependencies to ensure that the lockfile is up to date.
-                resolve_dependencies(
-                    &config,
-                    &config_path,
-                    &warg_config,
-                    &terminal,
-                    !self.dry_run,
-                )
-                .await?;
 
                 format!(
                     "dependency `{id}` with version `{version}`{dry_run}",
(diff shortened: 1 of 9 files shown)


crates/wit/src/lib.rs
```diff
@@ -273,6 +274,56 @@ struct PublishOptions<'a> {
...
+    }
+
+    if !config.categories.is_empty() {
+        metadata.set_categories(Some(config.categories.clone()));
+    }
+
+    metadata.set_description(config.description.clone());
+
+    // TODO: registry metadata should have keywords
+    // if !package.keywords.is_empty() {
```
fibonacci1729, line 290: Are `keywords` distinct from `categories`? They seem coupled in the original registry discussion.

crates/wit/src/commands/add.rs
```diff
@@ -127,16 +123,6 @@ impl AddCommand {
                     .dependencies
                     .insert(id.clone(), Dependency::Package(package));
 
-                // Resolve all dependencies to ensure that the lockfile is up to date.
-                resolve_dependencies(
```
fibonacci1729, line 131: I am sure i am missing something, but why is this removed?

fibonacci1729 (approved):

fibonacci1729:
LGTM! Just a couple question to educate myself.

crates/wit/src/lib.rs
peterhuene: `crates.io` has them as [distinct fields](https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field), where keywords is for describing the package and the category with a set of predefined [slugs](https://crates.io/category_slugs) to use.

I wasn't following the discussion around the metadata implementation closely, so it sounds like we should be populating this off of `keywords` instead of categories or perhaps both?

crates/wit/src/commands/add.rs
peterhuene: This was to fix a double-resolution that was taking place, once to resolve the package being added and then again to resolve all packages in an attempt to update the lock file.

Updating the lock file isn't strictly necessary from the command, so this was removed to improve the UX.

fibonacci1729: Awesome, makes complete sense. Thanks for the explanation.

crates/wit/src/lib.rs
peterhuene: I'm going to switch to populating the `categories` field from the cargo `keywords` as I believe that's more of the intention of this field; the predefined categories list from cargo doesn't map to what we'd want in a component registry anyway.

I've opened https://github.com/bytecodealliance/registry/issues/197 to track the confusion here.

peterhuene: I've decided to merge this as-is and wait for the resolution on the linked issue; will update `cargo-component` if things change.

PR #137, opened 2023-09-07
Use a default target path of `wit`.

### LINKED ISSUE #134: Could the `wit` folder be included by default without any configuration?
Similar to how Cargo supports `src/bin/*.rs` or `src/lib.rs` without configuration, the idea being that a conventional structure of a project doesn't require further configuration but config is always possible. To that effect would the `wit/*.wit` folder be loaded by default and made available for bindings?

peterhuene:
This PR makes `target.path` in the metadata default to a `wit` directory, if present.

This makes the `wit` directory a canonical directory, much like `src` is for source code.

Users may still override the path to find their local target definition.

Closes #134.

### DIFF
--- a/docs/design/registries.md
+++ b/docs/design/registries.md
@@ -321,9 +321,6 @@ version = "1.2.3"
 [package.metadata.component]
 package = "my-org:my-component"
 
-[package.metadata.component.target]
-path = "wit"
-
 [package.metadata.component.target.dependencies]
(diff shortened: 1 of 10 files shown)


peterhuene:
@alexcrichton this should help with command components that you just want to copy a wit into without having to modify `Cargo.toml` to point at a `wit` dir.

PR #140, opened 2023-09-18
Update WASI adapters to 2ad057d

jeffparsons:
This is for compatibility with the Wasmtime 13 release branch.

~~I have retained the previous adapters (134dddc) because I believe the plan is to support switching between them soon, and I imagine there would still be people wanting to use these older adapters even if running the latest version of cargo-component.~~


peterhuene:
I think we can delete the old adapters; generally we bump these every so often to whatever is latest out of Wasmtime, and wouldn't make use of a previous version of the adapter.

I'm thinking of implementing an option in `Cargo.toml` to point at the adapter to use:

```toml
[package.metadata.component]
adapter = "<path-to-adapter>"
```

If that setting is present, it'll use that, otherwise fall back to the built-in version which may or may not be current as right now it's still a manual process to update the embedded adapter.

peterhuene:
I should have the general purpose fix up today, by the way.

peterhuene (commented):
Looks good, but let's delete the previous adapters as part of this PR.

With my fix to add an `adapter` setting in `Cargo.toml`, users will be free to grab whatever version of the adapters they want out of the Wasmtime repo and we won't need the previous ones.

jeffparsons:
Fair enough; done!

I guess this will all be moot pretty soon anyway. :grinning:

peterhuene (approved):

peterhuene:
@jeffparsons Thanks again for fixing this. I should have that other PR up shortly.

PR #142, opened 2023-09-19
Add an `adapter` setting to `Cargo.toml`.

peterhuene:
This PR adds a `package.metadata.component.adapter` field to `Cargo.toml` which is the path to the WASI module adapter to use.

If present, the specified file is used to adapt WASI preview1 to preview2.

If not present, the built-in WASI module adapter is used instead.

The intention with this change is to allow users control over how WASI is adapted in cases where the built-in adapter is not the desired one to use.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,6 +18,7 @@ use metadata::ComponentMetadata;
 use registry::{PackageDependencyResolution, PackageResolutionMap};
 use semver::Version;
 use std::{
+    borrow::Cow,
     fs::{self, File},
     io::Read,
     path::{Path, PathBuf},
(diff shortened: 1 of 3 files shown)


peterhuene:
cc @jeffparsons

src/lib.rs
```diff
@@ -144,13 +145,18 @@ pub async fn run_cargo_command(
                     "debug"
                 });
 
-            for PackageComponentMetadata { package, .. } in packages {
+            for PackageComponentMetadata { package, metadata } in packages {
+                let metadata = match metadata {
+                    Some(metadata) => metadata,
+                    None => continue,
```
sunfishcode, line 151: This appears to be slightly different logic, skipping packages that don't have metadata, rather than proceeding to try to create components with them. Is that intended?

sunfishcode (commented):
Overall looks good, just one question:

jeffparsons:
Thanks, @peterhuene!

I expect to be periodically playing with fun new things on Wasmtime's main branch so I'll adopt this in my workflow and report/fix any issues I run into. (Although I'm about to go on parental leave so I'll be out of the world for a while soon!)

src/lib.rs
peterhuene: That's intended, as it shouldn't attempt to componentize the build output if `Cargo.toml` is devoid of a component section in the package metadata.

sunfishcode (approved):

Issue #94, opened 2023-07-01
Add `test` command to support running tests


ithinkicancode:
Currently there is no `test` command in `cargo-component`. The standard `cargo test` doesn't work because the standard tooling has no visibility to `bindings`, `exports` or any generated structs and enums. I'd think it's a good devX that devs can follow the same workflow just like when writing regular Rust apps.  Please point me to the right direction if I miss any docs wrt testing!

peterhuene:
Hi @ithinkicancode. You are correct that there currently isn't a `test` subcommand.

The main reason for that is that it would be nice to test the component itself in a wasm runtime rather than running it as a native executable via the default cargo test harness.

Some design needs to go into a proper implementation of that, though.

jakehemmerle:
i've been thinking about this too. one idea is that each function under test takes a component instance as input (ie an object that exports can be accessed through) and some custom `#[test]` macro that takes in custom testing parameters (eg. wasi or no wasi, runtime features, etc).

syntax is wrong but something like

```rust
#[component::test(
  wasi_enable = true,
)]
fn test_export_caller(component: Component) { // or whatever the proper type is
  assert!(component.exports.call_generate_rand_nums(32).is_ok()); // calls `generate_rand_nums()`, required us to enable wasi or error otherwise
}
```

ithinkicancode:
@jakehemmerle, do you have an example of instantiating a wasm blob (from the build) and writing tests against it? I'd be interested in trying out this route. :) Would appreciate it!

eduardomourar:
A temporary solution is to use the `CARGO_TARGET_<triple>_RUNNER` environment variable for that purpose. Assuming the latest wasmtime has been installed, you can run:
```console
CARGO_TARGET_WASM32_WASI_RUNNER="wasmtime --preview2 --wasm-features=component-model" cargo test
```

[Here](https://github.com/eduardomourar/wasm-web-shell/tree/6ec98d98d7490183e07e7152b150e5ad9ed3148a/packages/aws-cli) is a more complete example.

jakehemmerle:
> @jakehemmerle, do you have an example of instantiating a wasm blob (from the build) and writing tests against it? I'd be interested in trying out this route. :) Would appreciate it!

@ithinkicancode 

In this example, I instantiate a Wasm component with a Wasm runtime (in a wrapper) and then assert return values against its exported functions.

`ec_runtime/tests/runtime.rs`
```rust
[...]

PR #143, opened 2023-09-22
feat: support run, test and bench cargo commands

eduardomourar:
This will take a similar approach as used by `cargo-wasi` (source code [here](https://github.com/bytecodealliance/cargo-wasi/blob/e811d4889b516799d2f97813d073cec68b862adf/src/lib.rs#L114)) in order to expose the following cargo commands: `run`, `test` and `bench`.

The following steps will be done per action:

- `run`:  execute build command, componentize the output, and then feed what it componentized to the configured runner.
- `test` and `bench`: execute test/bench with `--no-run` argument, componentize the output, and then feed componentized wasm to the configured runner.

Related to: https://github.com/bytecodealliance/cargo-component/issues/94.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,7 +3,7 @@
 #![deny(missing_docs)]
 
 use crate::target::install_wasm32_wasi;
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{bail, Context, Result};
 use bindings::BindingsGenerator;
 use bytes::Bytes;
 use cargo_component_core::{
@@ -11,6 +11,7 @@ use cargo_component_core::{
     registry::create_client,
     terminal::{Colors, Terminal},
 };
+use cargo_config2::{PathAndArgs, TargetTripleRef};
 use cargo_metadata::{Metadata, MetadataCommand, Package};
 use config::{CargoArguments, CargoPackageSpec, Config};
 use lock::{acquire_lock_file_ro, acquire_lock_file_rw};
@@ -109,44 +110,50 @@ pub async fn run_cargo_command(
     let is_run = matches!(subcommand, Some("run") | Some("test") | Some("bench"));
 
     if is_run {
+        let cargo_config = cargo_config2::Config::load()?;
+
         // We check here before we actually build that a runtime is present.
         // We first check the CARGO_TARGET_WASM32_WASI_RUNNER environement
         // variable for a user-supplied runtime (path or executable) and use
         // the default, namely `wasmtime`, if it is not set.
-        let (wasi_runner, using_default) = env::var("CARGO_TARGET_WASM32_WASI_RUNNER")
+        let (runner, using_default) = cargo_config
+            .runner(TargetTripleRef::from("wasm32-wasi"))
+            .unwrap_or_default()
             .map(|runner_override| (runner_override, false))
-            .unwrap_or_else(|_| {
+            .unwrap_or_else(|| {
                 (
-                    "wasmtime -W component-model -S preview2 -S common".to_string(),
+                    PathAndArgs::new("wasmtime")
+                        .args(vec![
+                            "-W",
+                            "component-model",
+                            "-S",
+                            "preview2",
+                            "-S",
+                            "common",
+                        ])
+                        .to_owned(),
                     true,
                 )
             });
 
-        // Treat the wasi_runner variable as an exectable, followed by a whitespace-
+        // Treat the wasi_runner variable as an executable, followed by a whitespace-
         // separated list of arguments to the executable. This allows the user to
         // provide arguments which are passed to wasmtime without having to add more
         // command-line argument parsing to this crate.
-        let (wasi_runner, wasi_runner_extra_args) = {
-            let mut words = wasi_runner.split_whitespace();
-            let runner = words
-                .next()
-                .ok_or_else(|| anyhow!("$CARGO_TARGET_WASM32_WASI_RUNNER must not be empty"))?;
-            let extra_args = words.collect::<Vec<_>>();
-            (runner, extra_args)
-        };
+        let wasi_runner = runner.path.to_string_lossy().into_owned();
 
         if !using_default {
             // check if the override is either a valid path or command found on $PATH
-            if !(Path::new(&wasi_runner).exists() || which::which(wasi_runner).is_ok()) {
+            if !(runner.path.exists() || which::which(runner.path).is_ok()) {
                 bail!(
                     "failed to find `{}` (specified by $CARGO_TARGET_WASM32_WASI_RUNNER) \
                         on the filesytem or in $PATH, you'll want to fix the path or unset \
                         the $CARGO_TARGET_WASM32_WASI_RUNNER environment variable before \
                         running this command\n",
-                    &wasi_runner
+                    wasi_runner
                 );
             }
-        } else if which::which(wasi_runner).is_err() {
+        } else if which::which(runner.path).is_err() {
             let mut msg = format!(
                 "failed to find `{}` in $PATH, you'll want to \
                     install `{}` before running this command\n",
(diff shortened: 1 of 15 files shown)


eduardomourar:
@peterhuene, does this approach makes sense to you?

peterhuene:
Hi @eduardomourar, thanks for the PR!

Unfortunately, it isn't quite as simple as copying what `cargo wasi` does for using a runner to properly implement `cargo component run`.

The first problem is that componentization is expected to occur between the "build" and "run" stages of the "run" command; right now if you pass through to `cargo run` without a prior build, only a core module will get built before passing it through to the runner. This may appear to work for things like WASI commands because the core module is using WASI preview 1, which Wasmtime still supports from the CLI.

If we assume the first problem is solved, we run into the second problem: the component produced by `cargo-component` has (intentionally) abstract instance imports for any of its dependencies. To make the component suitable for running, we need to compose it with its dependencies prior to giving it to the runner. There's been work to automatically compose a component with its dependencies based on registry information and I'm currently working on a proper text format for `wasm-compose` to define custom compositions, but `cargo-component` will need to integrate with these before it produces runnable components.

To solve the first problem, `cargo component run` (and friends) will need to be implemented as proper subcommands rather than pass-throughs to `cargo` to allow for it to first spawn a build, componentize the output from `rustc`, and then spawn the runner.

Solving the second problem is much more involved and one of the reasons I've been holding off implementing these commands until a proper design is put in place.

eduardomourar:
@peterhuene, my particular use case is only to speed the development cycle while building individual components. You can find a sample here GitHub run [here](https://github.com/eduardomourar/wasm-web-shell/actions/runs/6316677281/job/17151804013#step:11:1). There you will see that I am running the test using this branch and the WASI Preview 2 functions are actually being invoked. It probably works because I don't have a composed component.

Is there anything I can change in this PR to get us in the expected direction? Maybe scope it down to include only the `test` subcommand.

peterhuene:
Your test is running `aws-cli` as a _core module_ and not as a component, and this is only working because the Wasmtime CLI, [until 5 days ago](https://github.com/bytecodealliance/wasmtime/pull/7056), supported `wasi:http` from a core module.

I suspect if you update your CI to use the latest Wasmtime, you'll get an error: `Cannot enable wasi-http for core wasm modules`.

As I mentioned earlier, the reason for this is that componentization of the build output occurs _after_ a `build` command. As a `test` or `run` command _implicitly_ executes the build command, there's no componentization occurring and you'll always get a core module being executed.

In addition to the problems I outlined above, `cargo test` will feed an output from deps path to the runner (in your CI example, that would be `/home/runner/work/wasm-web-shell/wasm-web-shell/target/wasm32-wasi/debug/deps/aws_cli-b90a0d2c7dc6f98b.wasm`), which `cargo component` currently does not attempt to componentize; `cargo component` only componentizes the target directory output (e.g. `target/wasm32-wasi/debug/foo.wasm`), which is a hardlink except on macOS afaik.

peterhuene:
That said, it is possible to implement a first-class `test` and `run` subcommand in `cargo-component` that does not delegate to `cargo test` or `cargo run`.

Those commands would do a build, componentizes the output, and then feed what it componentized to a configured runner.

That would probably be the first step in implementing this, upon which we can build integration of a composition step prior to executing the runner.

Implementing that

eduardomourar:
Thanks for the explanation. Apologies for the confusion on my part. I always forget that wasmtime will follow the component/core module path based on its content and not the arguments passed. I will check how complex it is to have the logic you suggested in this PR.

eduardomourar:
Ok, so finally got it working on unix. Just need to remove the hack to get the executable file.

peterhuene:
@eduardomourar thanks for all the work on this! I was away at a conference this week so haven't yet had time to review yet.

I'm still clawing my way out of a backlog, but I plan on reviewing this in full this coming Monday.

tests/bench.rs
```diff
@@ -0,0 +1,81 @@
...
+fn bench_recursive_fibonacci(b: &mut Bencher) {
+    b.iter(|| {
+        (0..5).map(fibonacci).collect::<Vec<u32>>()
+    })
+}"#,
+    )?;
+
+    project
+        .cargo_component("bench")
+        .env("RUSTUP_TOOLCHAIN", "nightly")
```
peterhuene, line 74: For my understanding, why the requirement on nightly?

src/lib.rs
```diff
@@ -100,82 +112,284 @@ pub async fn run_cargo_command(
...
     cmd.args(args);
 
-    let is_build = matches!(subcommand, Some("b") | Some("build") | Some("rustc"));
+    // We need to set the `color` parameter to "always", otherwise
+    // ANSI colors are being stripped out by default
+    if !build_args
+        .iter()
+        .any(|a| a == "--color" || a.starts_with("--color="))
+    {
+        cmd.arg("--color").arg("always");
```
peterhuene, line 131: As it appears that only `stderr` has colorized output when `--message-format json-render-diagnostics` is used, could we not use `.stdout(piped())` and get a reader which processes each line of JSON output, buffering only what wasm artifacts were built as it goes? Thus `stderr` would not be redirected, so we don't need to muck with `--color` and let it auto-detect a tty.

This would have the added benefit (and a complete deal breaker if not implemented) of not buffering the `stderr` output, making it look like `cargo-component` is hanging until the build completes and it blasts the entire output.

`cargo-component` is meant to look and feel like cargo itself, and if commands like `build` and `run` can't give immediate build progress output, then we need to find an alternative implementation.

peterhuene (commented):
Hi @eduardomourar, sorry for the delay.

I have some general comments about the implementation before going deeper into the review.

Generally, I do like the approach of using the JSON messages to detect what got built, which is much more robust than the current implementation.

But we simply cannot buffer cargo's output like this implementation, as it really hurts DX.

Would it be possible to explore using a pipe for stdout and process the JSON messages as they arrive?

peterhuene:
I will mention that using `--message-format` is exactly how `cargo-component` should have been detecting Rust toolchain outputs in the first place, so I really appreciate that you pushed forward with that approach.

tests/bench.rs
eduardomourar: As mentioned in the note [here](https://doc.rust-lang.org/cargo/commands/cargo-bench.html), the bench attribute can only used with nightly channel or with with some specific crates that bring the functionality into stable.

src/lib.rs
eduardomourar: I believe my first try was with a piped stdout and stderr but I ran into some complexity that I don't recall right now. I will make the modification and check if the approach you are suggesting will be more performant. I assume that performance/speed is what you were expecting to hurt DX, right?

eduardomourar:
By the way, I could not figure out how to ensure that only the current package is execute in a cargo workspace context.

src/lib.rs
```diff
@@ -100,82 +112,284 @@ pub async fn run_cargo_command(
...
+                                        "expected `{path}` to be a WebAssembly module",
+                                        path = path.display()
+                                    );
+                                }
+                                // Check for the module header version
+                                if bytes[4..8] == [0x01, 0x00, 0x00, 0x00] {
+                                    if is_test {
+                                        // To prevent the test entrypoint to clash with default WASI
+                                        // command export, we will rename it to something random
+                                        replace_export(bytes);
```
eduardomourar, line 323: The test/bench command will not work if the user overrides the adapter that is not of type command. I think the best approach would be to actually modify whatever adapter is resolved to so that the `run` export points to the test entrypoint. Unfortunately, I could not figure out how to do it. I did this hack of replacing the export `run` so at least it will work in the case of a reactor project with wasi cli that has no override.

src/lib.rs
peterhuene: Sorry I missed this question from two weeks ago (I was out last week too).

I'm not _as_ concerned about the performance (although that is very important) as I am about buffering the output of the spawned processes, which makes it appear like nothing is happening while cargo is running.

peterhuene (approved):
@eduardomourar thank you very much for your work on implementing these features!

I kept your approach mostly intact with my additional commits, except I piped stdout (when needed) and inherited stderr so that cargo both handles colors appropriately and output is displayed to users in "real time" (i.e. not buffering it).

The tests were modified slightly to rely on the cli world from the adapter, where applicable. This helped solve the conflicting `run` implementations.

PR #147, opened 2023-09-28
Add GitHub actions for publishing crates.

peterhuene:
This PR adds a GitHub action (based on a similar one from Wasmtime) for publishing the crates when new tags are created.

This also bumps the crate versions to 0.2.0, which will be the first versions of the crates published.

### DIFF
--- a/.github/workflows/publish-to-crates-io.yml
+++ b/.github/workflows/publish-to-crates-io.yml
@@ -1,0 +1,23 @@
+# The purpose of this workflow is to publish the workspace's crates
+# whenever a tag is created. This baiscally boils down to running
+# `scripts/publish.rs` at the right time.
+
+name: "Publish to crates.io"
+
+on:
+  push:
+    tags:
+    - 'v*'
+
+jobs:
+  publish:
+    if: github.repository == 'bytecodealliance/cargo-component'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: rustup update stable && rustup default stable
+    - run: |
+        rustc ci/publish.rs
+        ./publish publish
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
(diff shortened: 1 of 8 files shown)


ci/publish.rs
```diff
@@ -0,0 +1,406 @@
...
+use std::process::{Command, Stdio};
+use std::thread;
+use std::time::Duration;
+
+// note that this list must be topologically sorted by dependencies
+const CRATES_TO_PUBLISH: &[&str] = &[
+    "cargo-component-macro",
+    "cargo-component-bindings",
+    "cargo-component-core",
+    //"wit",
```
peterhuene, line 20: The `wit` tool is currently not published with the script as I need to talk to Dan to see if I can yoink his squatted crate name.

.github/workflows/publish-to-crates-io.yml
```diff
@@ -0,0 +1,23 @@
...
+    if: github.repository == 'bytecodealliance/cargo-component'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: rustup update stable && rustup default stable
+    - run: |
+        rustc ci/publish.rs
+        ./publish publish
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
morenol, line 23: the github ui syntax coloring makes me think that maybe it is safer to put `env:` first and then `run: |`

peterhuene: GitHub syntax coloring can be a little wonky at times, but ultimately this shouldn't matter.

For context, this CI job definition is a copy from other repros where it works as intended.

.github/workflows/publish-to-crates-io.yml
```diff
@@ -0,0 +1,23 @@
+# The purpose of this workflow is to publish the workspace's crates
+# whenever a tag is created. This baiscally boils down to running
```
fibonacci1729, line 2:
```suggestion
# whenever a tag is created. This basically boils down to running
```

fibonacci1729 (approved):

PR #150, opened 2023-09-29
feat(macro): Adds custom derive attributes

### LINKED ISSUE #148: Adding custom derive attributes
Wit bindgen recently merged support for adding custom derive attributes: https://github.com/bytecodealliance/wit-bindgen/pull/678. Is that something that can be exposed in `cargo component`? I'm fine to do the work, just need a little bit of direction about where it would belong

thomastaylor312:
This adds custom derive attributes to the generate macro. It also fixes a bug where package version was not appended to the hash map of exports

Closes #148

### DIFF
--- a/crates/macro/src/lib.rs
+++ b/crates/macro/src/lib.rs
@@ -4,7 +4,7 @@
 
 use heck::ToUpperCamelCase;
 use proc_macro2::{Span, TokenStream};
-use quote::quote;
+use quote::{quote, ToTokens};
 use std::{
     borrow::Cow,
     collections::HashMap,
@@ -22,8 +22,7 @@ use wit_bindgen_core::{
     wit_parser::{Resolve, TypeDefKind, WorldId, WorldItem, WorldKey},
     Files,
 };
-use wit_bindgen_rust::{ExportKey, Opts};
-use wit_bindgen_rust_lib::Ownership;
+use wit_bindgen_rust::{ExportKey, Opts, Ownership};
 use wit_component::DecodedWasm;
 
 fn implementor_path_str(path: &syn::Path) -> String {
@@ -61,6 +60,7 @@ fn implementor_path_str(path: &syn::Path) -> String {
 /// - `implementor`: The name of the type to implement world exports on.
 /// - `resources`: A map of resource names to resource implementor types.
 /// - `ownership`: The ownership model to use for resources.
+/// - `additional_derives`: Additional derive macro attributes to add to generated types
 ///
 /// # Examples
 ///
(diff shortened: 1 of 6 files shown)


thomastaylor312:
@peterhuene This is only draft because right now it is depending on a git dependency of wit-bindgen until 0.13 can be released. That release also combined two of the rust crates (see https://github.com/bytecodealliance/wit-bindgen/pull/683), so I had to make those changes here as well. I did test this manually with some projects I was using, but let me know if I missed anything

crates/macro/src/lib.rs
```diff
@@ -274,6 +285,10 @@ impl Config {
                     key.push_str(&package.name.name);
                     key.push('/');
                     key.push_str(interface.name.as_ref().expect("interface must have a name"));
+                    if let Some(version) = package.name.version.as_ref() {
```
peterhuene, line 288: Thanks for fixing this!

peterhuene (commented):
The implementation looks great!

I think we should probably add a build test that applies a drive and tests it (like a simple serialization test).

peterhuene:
Also, as this depends on an unreleased version of wit-bindgen, I'm going to hold off on merging until #147 is merged so that I can cut the first releases of the crates.

thomastaylor312:
> I think we should probably add a build test that applies a drive and tests it (like a simple serialization test).

@peterhuene I didn't add one because there is already a test in wit-bindgen that tests for this. Did you still want to add one here as a smoke test (just making sure it can compile with the `additional_derives` value set)?

thomastaylor312:
Ok, added a test anyway @peterhuene

peterhuene:
> @peterhuene I didn't add one because there is already a test in wit-bindgen that tests for this. Did you still want to add one here as a smoke test (just making sure it can compile with the additional_derives value set)?

Yeah, it'd be good to have the test to cover that the argument to `cargo_component_bindings::generate!` actually works, in addition to the test coverage in wit-bindgen to ensure the generated bindings are correct.

> Ok, added a test anyway @peterhuene

Thanks!

peterhuene (approved):

thomastaylor312:
I'll leave this in draft to avoid accidental merge until either

a) You've cut a version of this to release the crate
b) wit-bindgen releases a new tagged version

peterhuene:
We're now good to rebase this PR and change it from a draft; we don't need to wait for a wit-bindgen release.

thomastaylor312:
Ok, I'll rebase!

thomastaylor312:
@peterhuene This should be good to go

peterhuene:
@thomastaylor312 thanks!

PR #166, opened 2023-10-25
Reduce the size of published artifacts.

peterhuene:
This PR attempts to minimize the size of published artifacts by enabling LTO, aborting on panic, and stripping debug information.


peterhuene:
@alexcrichton would you mind reviewing? I figure for `cargo-component`, which has no unwinding, `abort` would be a fine default for the published artifacts; likewise the LTO cost isn't too great, so I enabled it.

alexcrichton (approved):

morenol (approved):

PR #167, opened 2023-10-27
Refactor bindings generation.

### LINKED ISSUE #139: Integration with rust-analyzer is broken for bindings
## Repro steps

1. `cargo component new repro`
2. `cd repro`
3. Run VS Code with rust-analyzer enabled and open `src/main.rs`.

rust-analyzer should trigger a check command that will generate a `target/bindings/repro/target.wasm` file containing the current target world for the project; the target is empty for command projects by default, so the generated `bindings` mod will be empty.

4. Create a `wit` directory.
5. Add a `test.wit` file with the following contents:

```wit
package repro:test

world repro {
    import test: func()
}
```

6. In VS Code, manually save `src/main.rs` to trigger a reparse.
7. In the `main` function in `src/main.rs`, type `bindings::`.

## Expected results

rust-analyzer should offer autocomplete suggestions for the imported function.

## Actual results

Nothing happens.

## Remarks

Right now, the `cargo_component_bindings::generate!` macro reads the target wasm file (which is very small and only contains type information) to generate bindings and emits a `include_bytes!` with the hopes that a dependency on the file will trigger a rebuild if it changes.

However, it appears that rust-analyzer doesn't respect changes to the file to reparse; it continues to use a cached evaluation of the proc-macro.

This may be related to https://github.com/rust-lang/rust-analyzer/issues/13668, but I can't say for sure.

## Workarounds

Restarting the rust-analyzer server after modifying the WIT files forces the re-evaluation of the macro with the latest bindings information.

peterhuene:
This PR alters how bindings generation works in `cargo-component`.

Previously, `cargo-component` would encode a target world for the `generate!` macro to decode. The macro would decode the target and feed it to `wit-bindgen` and expand to the generated bindings.

This approach had a few drawbacks:

* compilation times for projects increased as `wit-bindgen` had to be built from the component itself.
* rust-analyzer would not re-expand the `generate!` macro on WIT file changes as it does not add a source dependency via `include_bytes!`.
* "go to definition" doesn't work with rust-analyzer as the bindings code was entirely from the macro expansion.

To fix all three issues, `cargo-component` now generates the bindings source file itself and the macro simply expands to `include!` it.

This cuts the initial crate dependencies for a project from 50 to 13, speeding up the build (and our CI).

And as the `include_bytes!` is now an `include!`, rust-analyzer adds a source file dependency to re-expand the macro when any local WIT files change (on next reparse, that is).

However, this comes at the cost of being a breaking change:

* if the `cargo-component-bindings` crate version referenced from the project is mismatched from `cargo-component` following this change, the component project will not build as they won't agree on how to generate bindings.
* any users that provide arguments to the previous `generate!` macro must move those arguments into the `[package.metadata.component.bindings]` section in `Cargo.toml`.

Fixes #139.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -507,6 +507,7 @@ dependencies = [
  "wasm-metadata",
  "wasmparser 0.115.0",
  "wat",
+ "wit-bindgen-core",
  "wit-bindgen-rust",
  "wit-component",
  "wit-parser",
@@ -550,13 +551,9 @@ dependencies = [
 name = "cargo-component-macro"
 version = "0.3.1"
 dependencies = [
- "heck",
  "proc-macro2",
  "quote",
  "syn 2.0.38",
- "wit-bindgen-core",
- "wit-bindgen-rust",
- "wit-component",
 ]
 
 [[package]]
@@ -3843,7 +3840,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c7d92ce0ca6b6074059413a9581a637550c3a740581c854f9847ec293c8aed71"
 dependencies = [
  "bitflags 2.4.1",
- "wit-bindgen-rust-macro",
 ]
 
 [[package]]
@@ -3867,21 +3863,6 @@ dependencies = [
  "heck",
  "wasm-metadata",
  "wit-bindgen-core",
- "wit-component",
-]
-
-[[package]]
-name = "wit-bindgen-rust-macro"
-version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a91835ea4231da1fe7971679d505ba14be7826e192b6357f08465866ef482e08"
-dependencies = [
- "anyhow",
- "proc-macro2",
- "quote",
- "syn 2.0.38",
- "wit-bindgen-core",
- "wit-bindgen-rust",
  "wit-component",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ serde_json = { workspace = true }
 indexmap = { workspace = true }
 url = { workspace = true }
 wit-bindgen-rust = { workspace = true }
+wit-bindgen-core = { workspace = true }
 wit-parser = { workspace = true }
 wit-component = { workspace = true }
 wasm-metadata = { workspace = true }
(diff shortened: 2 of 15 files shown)


jeffparsons:
> * if the `cargo-component-bindings` crate version referenced from the project is mismatched from `cargo-component` following this change, the component project will not build as they won't agree on how to generate bindings.

Could `cargo-component` recognise this situation and advise the user?

(Future feature idea: offer to update it for you.)

peterhuene:
@jeffparsons it certainly can and I was planning on adding at least a warning if they're mismatched to this PR.

A small CLI command update the version in `Cargo.toml` for you (across the workspace) would be welcomed; perhaps a `cargo component upgrade` command that will run a `cargo install` to get the latest and update the versions in `Cargo.toml` accordingly? With perhaps a `cargo component upgrade --no-install` that just does the upgrade of the versions in `Cargo.toml` to whatever the current version is?

That would be as great follow-up PR.

elliottt (approved):
This looks good to me!

jeffparsons:
> A small CLI command to update the version in `Cargo.toml` for you (across the workspace) would be welcomed; perhaps a `cargo component upgrade` command that will run a `cargo install` to get the latest and update the versions in `Cargo.toml` accordingly? With perhaps a `cargo component upgrade --no-install` that just does the upgrade of the versions in `Cargo.toml` to whatever the current version is?
> 
> That would be a great follow-up PR.

I'll have a crack at this if I get a chance. No promises, though; my time is currently dominated by a toddler and a newborn! 😅

jeffparsons:
@peterhuene FYI, completely untested WIP: https://github.com/bytecodealliance/cargo-component/compare/main...jeffparsons:cargo-component:upgrade_subcommand?expand=1

I wouldn't waste your time reviewing at this point, because I haven't even run it. This is mostly just a heads-up that I did end up starting something. :blush:  

One detail I would like feedback on, though, is this design decision (copying a comment from the code):

```
        if !self.no_install {
            // Do the self-upgrade first, and then _unconditionally_ delegate
            // to whatever version of `cargo-component` is now at the same path as the
            // current executable.
            //
            // This avoids needing to query crates.io ourselves, scrape the version
            // from `cargo-component --version` etc.
            //
            // (We can't tell whether or not cargo-install actually installed anything
            // without scraping its output; it considers "already installed" as success.)
            upgrade_self()?;
            exec_cargo_component();
        }
```

Does that sound like a reasonable approach to you? Or do you think it's worth doing the work to avoid unnecessary invocations of `cargo-install` and `cargo-component`?

PR #175, opened 2023-10-29
Add `upgrade` subcommand

jeffparsons:
Installs the latest version of cargo-component and upgrade to the corresponding version of cargo-component-bindings.

Follow-up to https://github.com/bytecodealliance/cargo-component/pull/167#issuecomment-1783593614.

Before merge this needs at least:

- [ ] Review (duh)
- [x] Squash + meaningful commit message

In particular I would like to draw the reviewer's attention to:

- Design decision to unconditionally `cargo-install` and then ~~`exec`~~ delegate to either the existing binary or newly installed binary (unknown to us) for the rest of the job, as originally mentioned here: https://github.com/bytecodealliance/cargo-component/pull/167#issuecomment-1783943366.
- `cargo-install` part is currently completely untested. (Though it Works For Me™.) For now I've gated it behind `#[cfg(test)]` so I can still at least have the tests delegate to a child process to exercise as much of that code path as practicable. Any better ideas?
- I've added a boolean function parameter. Does that lead to [a paddlin'](https://simpsonswiki.com/wiki/That%27s_a_paddlin%27) around here? :sweat_smile: I'm happy to make it an enum with meaningful variant names if that's the preferred thing here.
- I haven't added a test that it actually updates all crates in the workspace. (Though again it Works For Me™.) Should I do that?

### DIFF
--- a/src/commands/upgrade.rs
+++ b/src/commands/upgrade.rs
@@ -1,0 +1,196 @@
+use crate::{load_metadata, Config, BINDINGS_CRATE_NAME};
+use anyhow::{Context, Result};
+use cargo_component_core::command::CommonOptions;
+use cargo_metadata::Metadata;
+use clap::Args;
+use std::{fs, path::PathBuf};
+use toml_edit::{value, Document};
+
+/// Install the latest version of cargo-component and upgrade to the
+/// corresponding version of cargo-component-bindings.
+#[derive(Args)]
+#[clap(disable_version_flag = true)]
+pub struct UpgradeCommand {
+    /// The common command options
+    #[clap(flatten)]
+    pub common: CommonOptions,
+
+    /// Don't actually write the Cargo.toml changes.
+    ///
+    /// Note that this will not prevent installing a new version of cargo-component itself;
+    /// if you want to do that, you must also specify the '--no-install' flag.
+    #[clap(long = "dry-run")]
+    pub dry_run: bool,
+
+    /// Path to Cargo.toml
+    #[clap(long = "manifest-path", value_name = "PATH")]
+    pub manifest_path: Option<PathBuf>,
+
+    /// Skip installing the latest version of cargo-component;
+    /// instead just upgrade cargo-component-bindings to match
+    /// the version currently running.
+    #[clap(long = "no-install")]
+    pub no_install: bool,
+}
+
+impl UpgradeCommand {
+    /// Executes the command.
+    pub async fn exec(self) -> Result<()> {
+        log::debug!("executing upgrade command");
+
+        if !self.no_install {
+            // Do the self-upgrade first, and then _unconditionally_ delegate
+            // to whatever version of `cargo-component` is now at the same path as the
+            // current executable.
+            //
+            // This avoids needing to query crates.io ourselves, scrape the version
+            // from `cargo-component --version` etc.
+            //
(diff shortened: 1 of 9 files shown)


jeffparsons:
@peterhuene I think this is basically what you suggested. PTAL? :pray:

jeffparsons:
Oh yeah, Windows. Forgot about that. 😅

I'll fix.

src/commands/upgrade.rs
```diff
@@ -0,0 +1,181 @@
+use crate::{load_metadata, Config, BINDINGS_CRATE_NAME};
+use anyhow::{Context, Result};
+use cargo_component_core::command::CommonOptions;
+use cargo_metadata::Metadata;
+use clap::Args;
+use std::{fs, os::unix::process::CommandExt, path::PathBuf};
```
jeffparsons, line 6: _Sigh_, I didn't even notice myself adding these. And I even thought to myself "huh, that's funny, I don't remember Windows supporting exec, but I guess I've been away for a while..." 😅

src/commands/upgrade.rs
```diff
@@ -0,0 +1,196 @@
...
+            log::debug!(
+                "Workspace package {} doesn't depend on cargo-component-bindings",
+                package.name
+            );
+            continue;
+        };
+
+        if bindings_dep.req == self_version {
+            config.terminal().status(
+                "Up-to-date",
```
peterhuene, line 135:
```suggestion
                "Skipping",
```

src/commands/upgrade.rs
```diff
@@ -0,0 +1,196 @@
...
+                format!(
+                    "Workspace package {} already specifies current version of cargo-component-bindings",
```
peterhuene, line 137:
```suggestion
                    "package `{}` as it already uses the current bindings crate version",
```

tests/upgrade.rs
```diff
@@ -0,0 +1,142 @@
...
+fn upgrade_single_crate_already_current_is_no_op() -> Result<()> {
+    let root = create_root()?;
+    let project = Project::with_root(&root, "component", "")?;
+
+    project
+        .cargo_component("upgrade")
+        .assert()
+        .success()
+        .stderr(contains(
+            "Up-to-date Workspace package component already specifies current version of cargo-component-bindings"
```
peterhuene, line 32: We'll need to update this expectation if we accept the above suggestions.

tests/upgrade.rs
```diff
@@ -0,0 +1,142 @@
...
+        Some(env!("CARGO_PKG_VERSION"))
+    );
+
+    // A repeated upgrade should recognize that there is no change required.
+    project
+        .cargo_component("upgrade")
+        .assert()
+        .success()
+        .stderr(contains(
+            "Up-to-date Workspace package component already specifies current version of cargo-component-bindings"
```
peterhuene, line 89: Same comment as above.

peterhuene (commented):
This looks really fantastic, thanks for taking this on!

I just have one minor suggestion to the message that gets printed when projects are up-to-date.

jeffparsons:
Good call; I've applied that change and rebased.

Thanks for the review! 💖

peterhuene (approved):
Thanks again for the PR!

PR #178, opened 2023-11-01
Update adapter to wasmtime 14.0.3

fibonacci1729:


peterhuene:
For my understand, as I am now generally bumping the adapter version only for major wasmtime releases and I don't see any wit changes from 14.0.0, is bumping the adapter necessary?

pchickey:
I just confirmed, there were no changes to the adapter source code, or the wit files, between 14.0.0 and 14.0.3. I am curious why the hash changed between those releases, did rustc change as well?

peterhuene:
The hash here is a wasmtime commit and not a hash of the adapter module, so I would expect differences between wasmtime versions.

fibonacci1729:
@peterhuene sorry for the noise! Thanks for confirming the process!

pchickey:
oh, right, I forgot we put the commit in there. Thanks

PR #180, opened 2023-11-06
Better error message when deserializing `RegistryPackage`

rylev:
Currently, when a component dependency is specified as a string it is assumed to be the version of the dependency on the registry. If, however, the user thinks that this could be path to a local wit definition, the error message is very confusing.

This changes adds a bit more context so hopefully the issue is clearer.


peterhuene (approved):
Thanks for fixing this!

PR #183, opened 2023-11-06
Cherry pick fix for 0.4.1 release.

peterhuene:


esoterra:
Why is `wit` being version bumped? I don't see any changes for for.

Also, is the release process documented somewhere for context?

peterhuene:
It's pretty much the same process as wasmtime's: there's a `ci/publish.rs` script used to bump the crate versions; once the PR to bump the versions is merged, then a tag is created to point at the merged commit. The act of creating the tag will trigger building the release artifacts.

Right now all the crates bump together to simplify things a bit; while technically this could be done to just the bindings crate, it's useful for at least `cargo-component` and `cargo-component-bindings` to match on version.

The `wit` tool shares the `cargo-component-core` dependency and technically only needs to bump if it bumps; we can probably refine this process a bit in the future to figure out exactly what crate needs bumping.

esoterra (approved):
LGTM

PR #187, opened 2023-11-08
Fix bindings version warning and upgrade command.

peterhuene:
This PR fixes the warning emitted when a mismatched bindings version crate is detected such that it no longer warns when the bindings crate version is compatible, but not the same version (i.e. `0.4.1` when cargo-component is `0.4.0`).

Likewise, the `upgrade` command no longer attempts to downgrade the version of the bindings crate when it would otherwise be compatible.

The upgrade tests no longer attempt the install, which require a long time to just verify that a `cargo install cargo-component` would succeed, which isn't terribly useful.

The warning and upgrade command now both detect when cargo-component is outdated and instruct the user appropriately.

### DIFF
--- a/src/commands/upgrade.rs
+++ b/src/commands/upgrade.rs
@@ -3,6 +3,7 @@ use anyhow::{Context, Result};
 use cargo_component_core::command::CommonOptions;
 use cargo_metadata::Metadata;
 use clap::Args;
+use semver::Version;
 use std::{fs, path::PathBuf};
 use toml_edit::{value, Document};
 
@@ -51,7 +52,6 @@ impl UpgradeCommand {
             //
             // Skip this in tests, but still delegate to a new instance of `cargo-component`
             // so that we can exercise as much of the flow as practicable.
-            #[cfg(not(test))]
             upgrade_self()?;
             run_cargo_component_and_exit();
         }
@@ -65,7 +65,6 @@ impl UpgradeCommand {
     }
 }
 
-#[cfg(not(test))]
 fn upgrade_self() -> Result<()> {
     log::debug!("running self-upgrade using cargo-install");
 
@@ -124,21 +123,27 @@ async fn upgrade_bindings(config: &Config, metadata: &Metadata, dry_run: bool) -
             .find(|dep| dep.name == "cargo-component-bindings")
         else {
             log::debug!(
-                "Workspace package {} doesn't depend on cargo-component-bindings",
-                package.name
+                "workspace package `{name}` doesn't depend on cargo-component-bindings",
+                name = package.name
             );
             continue;
         };
 
-        if bindings_dep.req == self_version {
-            config.terminal().status(
-                "Skipping",
-                format!(
-                    "package `{}` as it already uses the current bindings crate version",
-                    package.name
-                ),
-            )?;
-            continue;
+        // Treat the dependency req as a specific version
+        let version = bindings_dep.req.to_string();
+        let version = version.strip_prefix('^').unwrap_or(&version);
+        match version.parse::<Version>() {
+            Ok(v) if self_version.matches(&v) => {
+                config.terminal().status(
+                    "Skipping",
+                    format!(
+                        "package `{name}` as it already uses a compatible bindings crate version",
+                        name = package.name
+                    ),
+                )?;
+                continue;
+            }
+            _ => {}
         }
 
         let manifest_path = package.manifest_path.as_std_path();
@@ -167,7 +172,7 @@ async fn upgrade_bindings(config: &Config, metadata: &Metadata, dry_run: bool) -
                 format!(
                     "{path} from {from} to {to}",
                     path = manifest_path.display(),
-                    from = bindings_dep.req,
+                    from = version,
                     to = env!("CARGO_PKG_VERSION")
                 ),
             )?;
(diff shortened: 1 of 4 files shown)


jeffparsons:
> The upgrade tests no longer attempt the install, which require a long time to just verify that a `cargo install cargo-component` would succeed, which isn't terribly useful.

I thought the `#[cfg(not(test))]` achieved this? With the new approach of always passing in `--no-install` from the tests, I think that means delegation to a new process isn't tested anywhere.

src/commands/upgrade.rs
```diff
@@ -51,7 +52,6 @@ impl UpgradeCommand {
             //
             // Skip this in tests, but still delegate to a new instance of `cargo-component`
```
jeffparsons, line 53: Here's the rationale for the original approach. If this is changing, this bit of the comment should probably go, too.

src/lib.rs
```diff
@@ -220,27 +220,28 @@ pub fn load_metadata(
...
-                || dep
-                    .req
-                    .matches(&Version::parse(env!("CARGO_PKG_VERSION")).unwrap())
-            {
-                continue;
-            }
+                let req = VersionReq::parse(env!("CARGO_PKG_VERSION")).unwrap();
+                let v = dep.req.to_string();
+                match v.strip_prefix('^').unwrap_or(&v).parse::<Version>() {
+                    Ok(v) if req.matches(&v) => break,
```
jeffparsons, line 233: Does this work if the patch version difference is the other way around, e.g. if `cargo-component` was bumped to 0.4.2?

jeffparsons:
(Sorry if my comments here don't quite make sense; I'm posting from mobile in the middle of the night. Maybe I shouldn't do that...)

src/commands/upgrade.rs
peterhuene: Oh whoops, yeah, I'll remove the comment. Thanks!

src/lib.rs
peterhuene: You're right, this isn't quite adequate. What do you think of this approach instead?

* If major/minor differ:
  * If `cargo-component` version is _newer_ than the bindings crate, function like we do now (emit warning, allow upgrade of `Cargo.toml`)
  * If `cargo-component` version is _older_ than the bindings crate, emit an error saying cargo-component must be upgraded and do not allow upgrades to `Cargo.toml`.

jeffparsons: That sounds like it would do the trick!

peterhuene: Okay, I've pushed up a commit to hopefully address this and provide better feedback to users.

src/commands/upgrade.rs
```diff
@@ -124,22 +120,48 @@ async fn upgrade_bindings(config: &Config, metadata: &Metadata, dry_run: bool) -
...
-                    "package `{}` as it already uses the current bindings crate version",
-                    package.name
-                ),
-            )?;
-            continue;
-        }
+        let s = bindings_dep.req.to_string();
+        let version = match s.strip_prefix('^').unwrap_or(&s).parse::<Version>() {
+            Ok(v) => {
+                if this_version.major == v.major
```
jeffparsons, line 132: I'm trying to think about whether this makes sense, but I keep getting stuck. I think it's because I don't actually understand what the relationship is between different versions of cargo-component and the bindings crate, and therefore what — if anything — it means for them to be "semver-compatible with each other".

E.g. if cargo-component and the bindings both start at version 1.3, and then...

- cargo component has a new 1.4 release, is it allowed to rely on behaviour in the bindings crate that is only introduced in its corresponding version 1.4?
- the bindings crate has a new 1.4 release, is it allowed to assume that cargo-component will do something before or after it is built that is only implemented in cargo-component in its corresponding version 1.4?

I think what I'm slowly getting at is that I'm not sure "mutual semver compatibility" is actually meaningful.

One way out could be to always publish both in lockstep, and require an exact match on major.minor.

But again, because I don't understand what the relationship is _supposed_ to be, I might just be talking nonsense.

peterhuene: The bindings crate does very little (today, that is; previously it did the _actual_ bindings generation) other than include the bindings generated by `cargo-component` in the user's source file; to that end, they both need to agree on where the bindings are located.

I don't anticipate that relationship changing very often, but when it does, I think it would necessitate a bump of a major version (or minor when major is still 0).

Right now they publish with version bumps in lock-step, but only because our publish script isn't smart enough to bump only crates which have changed, and transitively, those that depend on them.

And also, since we're still at `0.y.z` of `cargo-component` itself, every release is pretty much a breaking change and they should probably be published lock-step until a 1.0.0.

Once there is a 1.0.0, though, I hope the bindings crate remains at 1.0.0 and only updates for major releases and any 1.y.z version of `cargo-component` can use it without having to warn users or bump the dependency in user's `Cargo.toml`s.

jeffparsons: Okay, I'm convinced. Thanks for the explanation! 😊

jeffparsons (approved):
This looks good to me! 🎉

I'm not a BA member or anything, though; I don't know if there are any rules for reviews that apply to this repo?

peterhuene:
As you are a previous contributor, I'm happy to rely on your review, given it is related to the feature you added previously.

I appreciate the review!

Issue #188, opened 2023-11-09
Build Failure with Inline WIT with `wit-bindgen`


tqwewe:
## Description

There's an inconsistency when building a project with inline WIT definitions using `cargo component build`. It fails under `cargo component` but succeeds when manually building with `cargo` and `wasm-tools` with `wit-bindgen` given the following `lib.rs`:

```rust
wit_bindgen::generate!({
    inline: r#"
        package component:guest;

        world example {
            export example: interface {
                hello-world: func() -> string;
            }
        }
    "#,
    exports: {
        "example": Component
    }
});

use exports::example::Guest;

struct Component;

impl Guest for Component {
    fn hello_world() -> String {
        "Hello, World!".to_string()
    }
}
```

### Manual Build with Cargo and wasm-tools

1. Build for `wasm32-wasi` target:
   ```bash
   cargo build --package guest --target wasm32-wasi
   ```
2. Create the component with wasm-tools:
   ```bash
   wasm-tools component new ./target/wasm32-wasi/debug/guest.wasm -o ./target/wasm32-wasi/debug/guest.wasm --adapt ./wasi_snapshot_preview1.wasm
   ```
3. Run the host package and observe successful `Hello, World!` message:
   ```bash
   cargo run --package host
   ```

### Build with Cargo Component

1. Clean previous builds:
   ```bash
   cargo clean
   ```
2. Build with `cargo component`:
   ```bash
   cargo component build --package guest
   ```
3. Observe failure when running the host package:
   ```bash
   cargo run --package host
   ``` 

## Expected Behavior

Both build methods should successfully compile and run the project, displaying `Hello, World!` regardless if we use a `wit` directory with `cargo_component_bindings` or `wit-bindgen`.

## Actual Behavior

The build succeeds with manual steps but fails with `cargo component`.

## Additional Information

Repository Link: [GitHub Repository](https://github.com/tqwewe/cargo-component-issue)

Is this intended behaviour?
I think it would be ideal if we can just use cargo component with wit-bindgen as an easy way of building wasm components, though it seems like we need to stick to the structure of having a `wit` directory along side our `src` directory.

peterhuene:
Could you provide the build error you're encountering?

peterhuene:
The guest project is missing the package metadata that marks it as a component project; currently `cargo-component` does not attempt to componentize just any `wasm` target outputs.

If you add `[package.metadata.component]` to your guest's `Cargo.toml`, does that workaround the issue?

[...]

PR #191, opened 2023-11-10
Use default component metadata when not present

tqwewe:
This PR makes the `[package.metadata.component]` section optional, using `ComponentSection::default()` if it's not present.

The thinking behind this, is that a user would never typically compile a regular wasm32-wasi module with `cargo component`. If they are using `cargo component build`, then they'd expect that the wasm module generated is a wasm component.

Closes https://github.com/bytecodealliance/cargo-component/issues/188

### DIFF
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -81,13 +81,6 @@ impl AddCommand {
                 )?,
             };
 
-        let metadata = metadata.with_context(|| {
-            format!(
-                "manifest `{path}` is not a WebAssembly component package",
-                path = package.manifest_path
-            )
-        })?;
-
         let id = match &self.id {
             Some(id) => id,
             None => &self.package.id,
--- a/src/commands/publish.rs
+++ b/src/commands/publish.rs
@@ -116,13 +116,7 @@ impl PublishCommand {
         )?];
 
         let package = packages[0].package;
-        let component_metadata = packages[0].metadata.as_ref().with_context(|| {
-            format!(
-                "package `{name}` is missing component metadata in manifest `{path}`",
-                name = package.name,
-                path = package.manifest_path
-            )
-        })?;
+        let component_metadata = &packages[0].metadata;
 
         let id = component_metadata.section.package.as_ref().with_context(|| {
             format!(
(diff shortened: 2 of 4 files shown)


peterhuene:
I think I'm generally onboard with this change, but we need to make sure that we only generate bindings only when the key is present, otherwise with this change I assume we'd generate empty bindings for every package in the workspace.

tqwewe:
> otherwise with this change I assume we'd generate empty bindings for every package in the workspace.

I see, I didn't realise thats how things were working internally.

I guess there's 2 possible solutions:
- **Could we only generate bindings for the crate thats being built somehow?**
  _Or would this have some issues of its own (such as if the crate being built, depends on another crate within the workspace)._
  
- **Is there any harm in generating bindings for all crates, even if they're not intended to be wasm components?**
  As mentioned, with this PR change, bindings might be generated for every package within the workspace.
  However, I wasn't aware of this when testing my PR, even when working in a host & guest workspace.
  Perhaps this would be fine, and wouldn't add any issues?

For the sake of simplicity, I still would love to avoid users having to mark every wasm component with `[package.metadata.component]`, especially for projects like mine where I don't want the user concerned with the concept of wasm components (as my crate handles this with its own macro built on top of wit-bindgen), and I'd just like them to be able to use `cargo component`, so they don't have to worry about `cargo build --target wasm32-wasi` and `wasm-tools component new ...` with the adapter.

tqwewe:
If you feel that you'd still like to go in the direction of marking each wasm component with the metadata, feel free to close this PR and the issue.

I've copied bits and pieces from the this codebase and implemented a `thalo build` command which does what i need (essentially wrapping `cargo rustc -p <name> --lib --target wasm32-wasi --crate-type=cdylib`)
https://github.com/thalo-rs/thalo/blob/main/crates/thalo_cli/src/cli/build.rs

peterhuene:
Hi @tqwewe. I just wanted to express my thanks for opening this issue and working on a fix.

Given the current design of bindings generation and the absence of the `package.component.metadata` section being relied upon to bypass bindings generation for non-component packages, I think it's probably best to close this PR for now.

I'm definitely open to revisiting this design in the future if there's substantial confusion from our users when the section is missing (i.e. they created their source with `cargo new`).

PR #206, opened 2024-01-11
Use current package with `add`, `publish` when in a workspace

tomasol:
Try to use current package in the `add` and `publish` command, when `-p` is not supplied and we are in a workspace.
Also isolate each test run. Previously all tests were sharing the same target/tests/Cargo.toml, which each test was modifying. This made order in which tests run significant and difficult to debug. Add test `two_projects_in_one_workspace_validate_add_from_path` which cretes two projects in the same workspace to make sure we modify the current package's manifest.

### DIFF
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -10,7 +10,7 @@ use cargo_component_core::{
     registry::{Dependency, DependencyResolution, DependencyResolver, RegistryPackage},
     VersionedPackageId,
 };
-use cargo_metadata::{Metadata, Package};
+use cargo_metadata::Package;
 use clap::Args;
 use semver::VersionReq;
 use std::{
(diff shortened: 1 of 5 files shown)


peterhuene:
Hi @tomasol.

Thanks very much for putting up this fix! I've put up #208 which also includes a fix for the tests, except it takes the approach of just using temporary directory for each test rather than stuffing them under `target/tests`.

Would you mind rebasing on that when it merges and update this PR with just the fix for the `add` subcommand in a workspace? Thanks a lot!

tomasol:
> Hi @tomasol.
> 
> Thanks very much for putting up this fix! I've put up #208 which also includes a fix for the tests, except it takes the approach of just using temporary directory for each test rather than stuffing them under `target/tests`.
> 
> Would you mind rebasing on that when it merges and update this PR with just the fix for the `add` subcommand in a workspace? Thanks a lot!

Hi @peterhuene , sure, I've made the adjustment, it was pretty straight-forward actually :slightly_smiling_face: 
I did not check all the commands which might benefit from this change however.

peterhuene (approved):
@tomasol sorry for the late review as your PR completely fell off my radar.

The changes look good and thanks for fixing this!

PR #209, opened 2024-01-25
Spawn `cargo add` in the `new` command

### LINKED ISSUE #204: Consider using `cargo add` in the `new` command
Currently `cargo component new` [generates](https://github.com/bytecodealliance/cargo-component/blob/205c6ee978d57b450cd1148a481e36c8bbb69c8c/src/commands/new.rs#L292) the dependency with the version fixed in the binary, e.g. `cargo-component-bindings = "0.6.0"` .
I would prefer having the same behavior as when calling `cargo add cargo-component-bindings`, generating `cargo-component-bindings.workspace = true` if I am inside a workspace and have version set in `[workspace.dependencies]`. If the version is not set, `cargo add` checks the latest version which is I believe preferable.

tomasol:
Fixes #204
`cargo component add` now spawns `cargo add cargo-component-bindings` . This has a few implications:
* It makes the command a bit slower as it needs to fetch the metadata from the network
* It uses the latest version by default.
* When in a workspace, and `cargo-component-bindings` is already defined as a workspace dependency, it uses the workspace version.

`cargo component upgrade` was not changed. I believe this command does too much and should be reconsidered.

Tests did not have to be changed, but checking a specific `cargo-component-bindings` version in tests might be problematic.

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -14,11 +14,10 @@ use std::{
     path::{Path, PathBuf},
     process::Command,
 };
-use toml_edit::{table, value, Array, Document, InlineTable, Item, Table, Value};
+use toml_edit::{table, value, Document, Item, Table, Value};
 use url::Url;
 
 const WIT_BINDGEN_CRATE: &str = "wit-bindgen";
-const WIT_BINDGEN_VERSION: &str = "0.16.0";
 
 fn escape_wit(s: &str) -> Cow<str> {
     match s {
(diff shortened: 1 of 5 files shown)


peterhuene:
Hi @tomasol. Thanks again for tackling this issue!

If we merge #212, the bindings crate has been removed entirely; however, it adds a dependency on `wit-bindgen` to pull in the runtime functions (disables the macro generation code, so it only has one additional transitive dependency).

So I think we should keep this change but update it for adding the dependency on `wit-bindgen` with `--no-default-features` and `--features realloc`.

That way we can remove the hardcoded version added by #212 and automatically pull it from a workspace if it's already there.

peterhuene:
I've pushed up the proposed change and fixed a bug where we weren't checking the returned status for a non-successful exit (the workspace related tests should have failed as it wasn't adding the dependency because the workspace was setup *before* the subprojects were created; just had to fix the ordering in the tests).

peterhuene (approved):
@tomasol thanks again for the fix!

tomasol:
@peterhuene Thanks for improving and merging the PR!

PR #211, opened 2024-01-25
Rename `--reactor` to `--lib` for `new` subcommand.

### LINKED ISSUE #198: Consider reverting `--reactor` to `--lib`
It seems as part of the large refactor in #101 the `--lib` flag was also renamed to `--reactor`. What's the reason for this change?

The term "reactor" suggests a component can somehow "react"? The term is commonly associated with reactive programming but it seems that's not what it should mean here.

The README admits the need to explain it means a library.

> A reactor component [..] is meant to be used as a library.

It somewhat makes sense if calling the component's exported functions makes it "react". But in that sense, a command component similarly "reacts" to its `run` function being called.

Is there a strong reason to replace the existing term "library" that's widely used across languages and ecosystems with a new term "reactor" which might be confused with an unrelated meaning? Why not call a library a library?

peterhuene:
This PR renames the `--reactor` option for the `new` subcommand to `--lib`.

It maintains an alias of `--reactor` for backwards compatibility.

Updated the tests to use `--lib` and added a test to ensure `--reactor` is still respected.

Closes #198.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,11 +206,12 @@ support in `cargo component` for adapting a _preview1_ module will be removed.
 
 ## Getting Started
 
-Use `cargo component new --reactor <name>` to create a new reactor component.
+Use `cargo component new --lib <name>` to create a new library (reactor) 
+component.
 
-A reactor component doesn't have a `run` (i.e. `main` in Rust) function
+A library component doesn't have a `run` (i.e. `main` in Rust) function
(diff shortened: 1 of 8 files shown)


sunfishcode (approved):
LGTM. As the person who publicly coined "reactor", I'm not attached to it, and think it makes sense to call things "libraries" when that's how developers are thinking about them.

Issue #196, opened 2023-11-21
Rust-analyzer doesn't integrate bindings in workspace


vwkd:
IntelliSense with rust-analyzer is broken in a workspace. The bindings in `target/bindings` are generated, but when expanding the macro recursively they aren't.

## Steps To Reproduce

1. Create workspace with one library crate. Minimal example: [cargo-component-rust-analyzer-workspace](https://github.com/vwkd/cargo-component-rust-analyzer-workspace).
1. Open repo in VSCode.
1. Confirm bindings are generated in [target/bindings/foo/bindings.rs](target/bindings/foo/bindings.rs#L6) automatically by rust-analyzer.
1. Confirm no IntelliSense when cursor is placed over `api::Guest` trait in [foo/src/lib.rs#L6](foo/src/lib.rs#L6).
1. Confirm running command "rust analyzer: Expand macro recursively at caret" gives empty `bindings` when cursor is placed over `cargo_component_bindings::generate!()` macro in [foo/src/lib.rs#L1](foo/src/lib.rs#L1).

## Versions

- VSCode `1.84.2`
- rust-analyzer extension `v0.3.1740`
- `cargo-component` package `0.4.1`

## Additional Information

The same issue occurs to me sometimes in a non-workspace library crate, but I can't reproduce it reliably. It seems like some random combination of restarting rust-analyzer, deleting the `target` directory, and changing `Cargo.toml` can fix it, and also break it again.

Continued from https://github.com/bytecodealliance/cargo-component/issues/139#issuecomment-1820609274

peterhuene:
From the output of RA:

```
File with cargo diagnostic not found in VFS: file not found: /Users/peterhuene/demo/target/bindings/proxy/bindings.rs
```

It seems, for whatever reason when loaded in a workspace, the VFS isn't populated with the bindings file, which I think means RA will give it an empty file to the proc-macro, ultimately resulting in an empty bindings macro.

PR #212, opened 2024-01-25
Remove `cargo_component_bindings` crate.

### LINKED ISSUE #210: Explain why the code is not generated in the `generate!()` macro
Looking at the implementation of `cargo_component_bindings::generate!()`, it does not generate any bindings, just `include!s` them, and the code is generated using `cargo component` command in [generate_bindings](https://github.com/bytecodealliance/cargo-component/blob/60e6880ef574b5b73f1c1edd88b810e18c2f6233/src/lib.rs#L508).

This is in conflict with the macro name, its documentation and with the readme:
> Like wit-bindgen, cargo component uses a procedural macro to generate bindings. However, bindings are generated based on the resolved dependencies from Cargo.toml rather than parsing a local definition of the component's interface.

This also brings additional problems like the currently broken code completion ( #196) and the need to reconfigure rust-analyzer.

What is the reasoning behind the change? Should the macro be renamed to clarify its responsibility?

peterhuene:
This PR is a breaking change to `cargo-component` with the intention of improving component build time (now down to two transitive crate dependencies) and get `rust-analyzer` fully working with bindings.

It removes the `cargo_component_bindings` and `cargo_component_macro` crates. Instead of using a macro for including the generated bindings, `cargo-component` now directly writes to a `src/bindings.rs` file which is included as a submodule in `main.rs` or `lib.rs`.

Additionally, the `update` subcommand was removed as it is no longer necessary since there is no `cargo_component_bindings` crate referenced in `Cargo.toml`; the functionality to update `cargo-component` itself can simply be replaced with a `cargo install` or `cargo binstall`.

Closes #210 and #196.


peterhuene:
Putting this up as a proposal for removing the `cargo_component_bindings` crate in favor of just having `cargo-component` generate the bindings source directly into the component projects.

This has the benefit of cutting down on the number of dependencies from user's component packages and also getting `rust-analyzer` to work in workspaces as there's no longer any macro that reads a file in a `target` subdirectory.

fibonacci1729 (approved):

peterhuene:
In the future, I'd love for `wit-bindgen` to be capable of (optionally) generating the runtime support functions into the bindings source itself so that we can remove the `wit-bindgen` dependency from user components, but we can live with it for now.

DougAnderson444:
Hi @peterhuene,

I'm following this PR and looking at updating to `v0.7.0`, but ran into issues until I realzied that `Cargo.toml` also has to change to:

```toml
[dependencies]
wit-bindgen = "0.16.0"

# Now this is a build dep?
[build-dependencies]
cargo-component= "0.7.0"
```

Is this right? Seems to work. The readme might need to be updated too (possibly a separate issue?)

DougAnderson444:
Ah, it's just an upgrading issue really since I already had an existing repo. I tried `cargo component new --lib` and it confirmed we don't need the `build-dep`

PR #221, opened 2024-01-26
Add a `package.metadata.component.proxy` setting.

peterhuene:
This PR adds a `package.metadata.component.proxy` setting to `Cargo.toml` that controls whether or not to use the `wasi:http/proxy` specific adapter.

This option can be set when creating a project with `cargo component new --lib --proxy`.

The `wasi:http/proxy` adapter is now built-in to `cargo-component`; it adapts from preview1 in a way that ensures the resulting component conforms to the `wasi:http/proxy` world.

Additionally, the `--command` option has been renamed to `--bin`, with an alias for `--command` for backwards compatibility.


sunfishcode (approved):
Cool, I like the --proxy argument to cargo component new that sets this variable automatically.

PR #223, opened 2024-01-28
tests fix

macovedj:
Currently when building off of main, tests are failing out of the box.

Some tests were failing because they are using flags from before `wasmtime` enabled the component model by default, so those flags are no longer being used.

Others were matching against information output by running `cargo component metadata` and seemed to not expect json output, so there are changes for that too.


peterhuene (approved):
👍

PR #224, opened 2024-01-29
Anticipate packages

macovedj:
Today, one can publish a `cargo-component` generated component to a registry and then add it as a dependency in another  `cargo-component` project.  However, the consuming component's wit will not refer to a foreign package if an interface is imported, but instead will define that interface inline.  The example below should help make this clear.

Given a simple component with the following wit:

```
package component:hello;

interface greeter {
  hello-world: func() -> string;
}
/// An example world for the component to target.
world example {
  import greeter;
  export greeter;
}
```

That is added via `cargo-component add` to another component, we can reference it in the project that is added to like so:

```rs
use bindings::hello_greeter::hello_world;
use bindings::Guest;

struct Component;

impl Guest for Component {
    /// Say hello!
    fn goodbye_world() -> String {
        hello_world()
    }
}
```

However, when printing out the wit of the compiled component, the `greeter` interface appears as follows:

```
package root:component;

world root {
  import hello-greeter: interface {
    hello-world: func() -> string;
  }
  import wasi:cli/environment@0.2.0;
  import wasi:cli/exit@0.2.0;
  import wasi:io/error@0.2.0;
  import wasi:io/streams@0.2.0;
  import wasi:cli/stdin@0.2.0;
  import wasi:cli/stdout@0.2.0;
  import wasi:cli/stderr@0.2.0;
  import wasi:clocks/wall-clock@0.2.0;
  import wasi:filesystem/types@0.2.0;
  import wasi:filesystem/preopens@0.2.0;

  export goodbye-world: func() -> string;
}
```

We can see that it is not surfaced to the wit that this interface is defined by the dependency.  The use case I specifically am motivated by is in is docs gen so that a UI can link to the package where the type is actually defined.

These changes will produce a binary with the following wit

```
world root {
  import wasi:hello/greeter;
  import wasi:cli/environment@0.2.0;
  import wasi:cli/exit@0.2.0;
  import wasi:io/error@0.2.0;
  import wasi:io/streams@0.2.0;
  import wasi:cli/stdin@0.2.0;
  import wasi:cli/stdout@0.2.0;
  import wasi:cli/stderr@0.2.0;
  import wasi:clocks/wall-clock@0.2.0;
  import wasi:filesystem/types@0.2.0;
  import wasi:filesystem/preopens@0.2.0;

  export goodbye-world: func() -> string;
}
```

For the motivating use case, this would surface sufficient information for robust docgen.

Note that as it stands, in order for these changes to function properly would also require that the `wit-bindgen` branch referenced gets merged, which has the side effect of prepending the namespace to the module exposed in rust bindings, and the rust import would then look like so

```rs
use bindings::<namespace>_hello_greeter::hello_world;
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -545,7 +545,7 @@ dependencies = [
  "which 6.0.0",
  "wit-bindgen-core",
  "wit-bindgen-rust",
- "wit-component 0.20.0",
+ "wit-component",
  "wit-parser",
 ]
 
@@ -571,7 +571,7 @@ dependencies = [
  "warg-crypto",
  "warg-protocol",
  "windows-sys 0.52.0",
- "wit-component 0.20.0",
+ "wit-component",
  "wit-parser",
 ]
 
@@ -3654,15 +3654,6 @@ checksum = "4d91413b1c31d7539ba5ef2451af3f0b833a005eb27a631cec32bc0635a8602b"
 
 [[package]]
 name = "wasm-encoder"
-version = "0.38.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ad2b51884de9c7f4fe2fd1043fccb8dcad4b1e29558146ee57a144d15779f3f"
-dependencies = [
- "leb128",
-]
-
-[[package]]
-name = "wasm-encoder"
 version = "0.40.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d162eb64168969ae90e8668ca0593b0e47667e315aa08e717a9c9574d700d826"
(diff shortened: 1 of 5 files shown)


macovedj:
I commented out a test that I'm still trying to get to work that would verify we get the expected results when publishing components to a registry.  But I have tested against real components and it works.

I also brought over the test changes from [here](https://github.com/bytecodealliance/cargo-component/pull/223), and if these changes end up feeling like the correct approach, I can just go ahead and close the other PR.  

Can open up the `wit-bindgen` PR also once it's confirmed that this approach is fine and/or get good tests pushed.

PR #243, opened 2024-03-01
Bump dependencies to latest.

peterhuene:
This includes a breaking change from `wit-bindgen` that reintroduces the generated `export!` macro.

The settings under `[package.metadata.component.bindings]` have changed as a result:

* The `implementor` setting has been removed as it is now explicitly specified via the `export!` macro.
* The `resources` setting has been removed as it is now specified via trait associated types.
* Introduced the `format` setting (defaults to `true`) to control whether or not the generated bindings are run through `rustfmt`.

Additionally, `cargo component new` has been updated to use `wit-bindgen-rt` as a dependency rather than a feature-gated `wit-bindgen` dependency; this is also a breaking change.

These breaking changes will be documented heavily in the release notes.


elliottt (approved):
Looks great!

PR #251, opened 2024-03-05
rely on warg-credentials for keyring management

macovedj:
leverage warg-credentials for keyring management


Cargo.toml
```diff
@@ -98,14 +102,15 @@ rpassword = "7.3.1"
...
+secrecy = "0.8.0"
 wit-bindgen-rust = "0.21.0"
 wit-bindgen-core = "0.21.0"
 tempfile = "3.10.1"
 assert_cmd = "2.0.14"
 predicates = "3.1.0"
 wasmparser = "0.201.0"
 wat = "1.201.0"
-warg-server = "0.3.0"
+warg-server = "0.4.1"
```
peterhuene, line 113: Nit (and not your issue): do you mind moving this up with the other warg-related deps? It's only a dev dependency, but cargo doesn't distinguish between dev and non-dev deps at the workspace level. Thanks!

Cargo.toml
```diff
@@ -45,13 +45,16 @@ libc = { workspace = true }
...
+warg-credentials = { workspace = true }
 p256 = { workspace = true }
 rand_core = { workspace = true }
 rpassword = { workspace = true }
 futures = { workspace = true }
 bytes = { workspace = true }
 which = { workspace = true }
 shell-escape = "0.1.5"
+secrecy = { workspace = true }
+keyring = { workspace = true }
```
peterhuene, line 57:
```suggestion
```

peterhuene: We can delete it from `[workspace.dependencies]` too.

crates/core/Cargo.toml
peterhuene, line 1: Need to delete the `keyring` dependency in this file.

crates/wit/Cargo.toml
```diff
@@ -30,6 +31,7 @@ indexmap = { workspace = true }
 wit-parser = { workspace = true }
 wit-component = { workspace = true }
 wasm-metadata = { workspace = true }
+keyring = { workspace = true }
```
peterhuene, line 34:
```suggestion
```

src/commands/key.rs
```diff
@@ -135,30 +112,36 @@ impl KeyNewCommand {
...
-    pub async fn exec(self, config: &Config) -> Result<()> {
+    pub async fn exec(self, config: &mut Config) -> Result<()> {
         let key = PrivateKey::decode(
             rpassword::prompt_password("input signing key (expected format is `<alg>:<base64>`): ")
                 .context("failed to read signing key")?,
         )
         .context("signing key is not in the correct format")?;
 
-        set_signing_key(&self.url, &self.key_name, &key)?;
+        // let key = PrivateKey::decode(key).context("signing key is not in the correct format")?;
```
peterhuene, line 129:
```suggestion
```

peterhuene (commented):
Looks great, thanks for making these changes!

Just some minor stuff to correct.

PR #265, opened 2024-03-22
Implement `wit pull`

lann:
This implements a `wit pull` subcommand which can be used to update local e.g. `wit/deps/*.wit` from a registry.

```console
$ wit pull wasi:io --create-dirs
   Resolving package wasi:io
  Downloaded release wasi:io@0.2.0
       Wrote package wasi:io@0.2.0 to 'wit/deps/wasi-io@0.2.0.wit'

$ head -n1 wit/deps/wasi-io@0.2.0.wit
package wasi:io@0.2.0;

$ wit pull wasi:http
   Resolving package wasi:http
  Downloaded release wasi:http@0.2.0
       Wrote package wasi:clocks@0.2.0 to 'wit/deps/wasi-clocks@0.2.0.wit'
       Wrote package wasi:random@0.2.0 to 'wit/deps/wasi-random@0.2.0.wit'
       Wrote package wasi:cli@0.2.0 to 'wit/deps/wasi-cli@0.2.0.wit'
       Wrote package wasi:http@0.2.0 to 'wit/deps/wasi-http@0.2.0.wit'
```

There are two modes of operation:

- When given specific name(s) it will pull those registry package(s) and unpack all (not existing) WIT packages into `wit/deps/<namespace>-<name>@<version>.wit`
- When given no name(s) it will check the packages under `wit/` for missing dependencies and attempt to pull them like above

Both of these can operate whether or not `wit_parser::Resolve::push_dir` can currently parse the `wit/` directory. If it _can_ parse, the existing dependency checks will just be more accurate (capturing e.g. `wit/deps/some-package.wasm`). This ought to be more consistent but really needs some refactoring in `wit-parser` first.

### DIFF
--- a/crates/wit/src/commands/pull.rs
+++ b/crates/wit/src/commands/pull.rs
@@ -23,6 +23,8 @@ pub struct PullCommand {
     pub common: CommonOptions,
 
     /// Use the specified registry name as the default when pulling package(s).
+    /// Note: Currently 'wasi:' packages will be pulled from
+    /// 'bytecodealliance.org' regardless of this flag.
     #[clap(long, value_name = "REGISTRY")]
     pub registry: Option<String>,
 
@@ -111,8 +113,7 @@ impl PullCommand {
                     };
                     terminal.status("Downloaded", format!("release {release_pkg}"))?;
 
-                    let version = root_pkg.name.version.clone();
-                    if let Some(wit_version) = &version {
+                    if let Some(wit_version) = &root_pkg.name.version {
                         let release_version = &release.version;
                         if wit_version != release_version {
                             terminal.warn(format!("Release version {release_version} doesn't match WIT package version {wit_version}"))?;
--- a/crates/wit/src/commands/pull.rs
+++ b/crates/wit/src/commands/pull.rs
@@ -205,7 +205,6 @@ impl PullCommand {
     }
 }
 
-#[derive(Debug)]
 struct PackagesState {
     // Packages currently present in the wit dir
     present: BTreeSet<PackageName>,
@@ -234,6 +233,11 @@ impl PackagesState {
         // Root package
         match UnresolvedPackage::parse_dir(wit_dir) {
             Ok(pkg) => {
+                log::debug!(
+                    "Parsed root package '{name}' deps = {deps:?}",
+                    name = pkg.name,
+                    deps = debug_pkg_names(pkg.foreign_deps.keys()),
+                );
                 present.insert(pkg.name);
                 deps.extend(pkg.foreign_deps.into_keys());
             }
@@ -258,6 +262,11 @@ impl PackagesState {
                 }
                 .with_context(|| format!("Error parsing {path:?}"))?;
 
+                log::debug!(
+                    "Parsed deps package '{name}' deps = {deps:?}",
+                    name = pkg.name,
+                    deps = debug_pkg_names(pkg.foreign_deps.keys()),
+                );
                 present.insert(pkg.name);
                 deps.extend(pkg.foreign_deps.into_keys());
             }
@@ -289,4 +298,20 @@ impl PackagesState {
         }
         false
     }
+}
+
+impl std::fmt::Debug for PackagesState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PackagesState")
+            .field("present", &debug_pkg_names(&self.present))
+            .field("deps", &debug_pkg_names(&self.deps))
+            .finish()
+    }
+}
+
+fn debug_pkg_names<'a>(names: impl IntoIterator<Item = &'a PackageName>) -> Vec<String> {
+    names
+        .into_iter()
+        .map(ToString::to_string)
+        .collect::<Vec<_>>()
 }
--- a/crates/wit/src/commands/pull.rs
+++ b/crates/wit/src/commands/pull.rs
@@ -1,0 +1,183 @@
+use std::{
+    io::Write,
+    path::{Path, PathBuf},
+};
+
+use anyhow::{bail, Context, Result};
+use clap::Args;
+
+use cargo_component_core::command::CommonOptions;
+use futures::TryStreamExt;
+use semver::Version;
+use tokio_util::io::{StreamReader, SyncIoBridge};
+use warg_loader::PackageRef;
+use wit_parser::UnresolvedPackage;
+
+/// Pull WIT package(s) to a local "deps" directory.
+#[derive(Args)]
+#[clap(disable_version_flag = true)]
+pub struct PullCommand {
+    /// The common command options.
+    #[clap(flatten)]
+    pub common: CommonOptions,
+
+    /// Use the specified registry name when pulling the package(s).
+    #[clap(long, value_name = "REGISTRY")]
+    pub registry: Option<String>,
+
+    /// Update the specified directory WIT "root" directory. Dependencies will
+    /// be written to e.g. "<wit-dir>/deps/<namespace>.<package>.wit".
+    #[clap(long, value_name = "WIT_DIR", default_value = "wit")]
+    pub wit_dir: PathBuf,
+
+    /// Create "<wit-dir>" and "<wit-dir>/deps" directories as needed.
+    #[clap(long)]
+    pub create_dirs: bool,
+
+    /// Pull the packages specified. If empty, the list of packages to pull will
+    /// be parsed from "<wit-dir>/deps/*.wit".
+    #[clap(value_name = "PACKAGE")]
+    pub packages: Vec<PackageRef>,
+}
+
+impl PullCommand {
+    /// Executes the command.
+    pub async fn exec(self) -> Result<()> {
+        log::debug!("executing pull command");
+
+        let terminal = self.common.new_terminal();
+
+        let deps_dir = self.wit_dir.join("deps");
+
+        let deps = if self.packages.is_empty() {
+            let deps = if deps_dir.exists() {
+                Dep::parse_deps_files(&deps_dir)
+                    .with_context(|| format!("Couldn't read existing deps from {deps_dir:?}"))?
+            } else {
+                vec![]
+            };
+            if deps.is_empty() {
+                terminal.warn(format!(
+                    "No deps found at {glob:?}; nothing to pull.",
+                    glob = deps_dir.join("*.wit")
+                ))?;
+                return Ok(());
+            }
+            deps
+        } else {
+            self.packages
+                .into_iter()
+                .map(|pkg| Dep::specific(pkg, &deps_dir))
+                .collect::<Result<Vec<_>>>()?
+        };
+
+        let mut client = match warg_loader::Client::from_default_config_file()? {
+            Some(client) => client,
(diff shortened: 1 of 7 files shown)


lann:
I'll be on vacation through Mar 26. Someone else is welcome to pick this up if desired.

lann:
One open question is what to do with the `warg-loader` crate. At this point it _could_ make sense to merge that into this repo.

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// Use the specified registry name when pulling the package(s).
+    #[clap(long, value_name = "REGISTRY")]
+    pub registry: Option<String>,
+
+    /// Update the specified directory WIT "root" directory. Dependencies will
+    /// be written to e.g. "<wit-dir>/deps/<namespace>.<package>.wit".
+    #[clap(long, value_name = "WIT_DIR", default_value = "wit")]
+    pub wit_dir: PathBuf,
+
+    /// Create "<wit-dir>" and "<wit-dir>/deps" directories if needed.
```
rylev, line 34: I guess this would create wit-dir if it doesn't exist, but as I indicate below, I think that's an error condition and this should only be used to create the `deps` subdirectory if it doesn't exist.

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,273 @@
...
+    #[clap(long)]
+    pub create_dirs: bool,
```
rylev, line 36: I wonder if creating the directory should be the default and the user can opt out of it...

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,273 @@
...
+                let path = entry.path();
+                let pkg = if entry.file_type()?.is_dir() {
+                    UnresolvedPackage::parse_dir(&path)
+                } else if path.extension().unwrap_or_default() == "wit" {
+                    UnresolvedPackage::parse_file(&path)
+                } else {
+                    continue;
+                }
+                .with_context(|| format!("Error parsing {path:?}"))?;
+
```
rylev, line 174: Can we debug log here both `pkg.name` and `pkg.foreign_deps.keys()`?

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,273 @@
...
+            }
+            Err(err) => log::debug!("Couldn't resolve packages: {err:#}"),
+        }
+
+        // Approximate the Resolve::push_dir process
+        // TODO: Try to expose this logic from wit-parser instead
+        let mut pkgs = match UnresolvedPackage::parse_dir(&self.wit_dir) {
+            Ok(pkg) => BTreeSet::from([pkg.name]),
+            Err(err) => {
+                log::debug!("Couldn't parse root package: {err:?}");
```
rylev, line 156: This seems like something we shouldn't silently ignore - but either error on or emit a warning.

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,273 @@
...
+                    Default::default(),
+                ))
+            }
+            Err(err) => log::debug!("Couldn't resolve packages: {err:#}"),
+        }
+
+        // Approximate the Resolve::push_dir process
+        // TODO: Try to expose this logic from wit-parser instead
+        let mut pkgs = match UnresolvedPackage::parse_dir(&self.wit_dir) {
+            Ok(pkg) => BTreeSet::from([pkg.name]),
```
rylev, line 154: The `pkg`'s foreign deps should be added to `deps` here. Otherwise, we only fetch dependencies of dependencies and not the main package's direct dependencies.

rylev (changes requested):
Seems like there's an issue with fetching direct dependencies that we should fix before merging, but overall this looks good. 

I am getting errors when trying this out myself, presumably because I don't have access to the `bytecodealliance.org` registry. I think we'll want to figure out the defaults around registries as well before merging.

crates/wit/src/commands/pull.rs
lann: Ah yeah I lost track of that while refactoring :facepalm:

lann: Fixed and refactored a lot of this logic into a `struct PackagesState` like I should have started with :roll_eyes::  546b092

crates/wit/src/commands/pull.rs
lann: I was thinking about `--create-dirs` in terms of initing a new project, but I suppose the chance of accidentally doing this in the wrong parent dir probably outweighs the tiny inconvenience of creating one dir.

I think I'll get rid of the option and just create the `deps` dir unconditionally when needed.

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,273 @@
...
+        };
+
+        if packages.is_empty() {
+            terminal.status("Finished", "no missing packages; nothing to do")?;
+            return Ok(());
+        }
+
+        let mut client = {
+            let mut config = ClientConfig::default();
+            config.namespace_registry("wasi", "bytecodealliance.org");
```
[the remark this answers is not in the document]
lann: Now I'm not sure what the best UX is here. If you specify `--registry` you definitely expect at least some package to come from there, but then should it fall back to the config file / defaults otherwise? :thinking:

crates/wit/src/commands/pull.rs
lann: c63c145

crates/wit/src/commands/pull.rs
lann: I added a warning if the root package is unparsable and you didn't pass any explicit packages.

I think it would be somewhat common to have an unparsable root exactly when you are running this command, e.g. bootstrapping a new project or in the middle of making changes that need new dependencies.

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        // If Resolve::push_dir succeeds there are no unresolved deps
+        match resolve.push_dir(&self.wit_dir) {
+            Ok(_) => {
+                return Ok((
+                    resolve.package_names.into_keys().collect(),
+                    Default::default(),
+                ))
+            }
+            Err(err) => log::debug!("Couldn't resolve packages: {err:#}"),
```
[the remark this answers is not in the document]
lann: https://github.com/bytecodealliance/cargo-component/pull/265#discussion_r1549822010

crates/wit/src/commands/pull.rs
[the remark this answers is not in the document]
lann: I'm going to punt a bit and do the thing that is currently easy, which is for `--registry` to only set the _default_ registry which will still be overridden by other configured namespace mappings. I don't think this is ideal but should work most of the time. a0e0e51

lann:
> I am getting errors when trying this out myself, presumably because I don't have access to the bytecodealliance.org registry.

The BA registry should be public, e.g. https://github.com/orgs/bytecodealliance/packages/container/package/warg%2Fwasi%2Fio

rylev:
This is the error I'm seeing when running the command against a wit world that uses a `wasi:http` dependency.

```
error: OCI error: Authentication failure: {"errors":[{"code":"DENIED","message":"denied"}]}


Caused by:
    Authentication failure: {"errors":[{"code":"DENIED","message":"denied"}]}
```

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,324 @@
...
+
+                    let name = &pkg.name;
+                    let release_pkg = PackageName {
+                        namespace: name.namespace().to_string(),
+                        name: name.name().to_string(),
+                        version: Some(release.version.clone()),
+                    };
+                    terminal.status("Downloaded", format!("release {release_pkg}"))?;
+
+                    let version = root_pkg.name.version.clone();
```
rylev, line 114: Nit: is this clone necessary?

crates/wit/src/commands/pull.rs
```diff
@@ -0,0 +1,324 @@
...
+                    if let Some(wit_version) = &version {
+                        let release_version = &release.version;
+                        if wit_version != release_version {
+                            terminal.warn(format!("Release version {release_version} doesn't match WIT package version {wit_version}"))?;
```
rylev, line 118: I'm curious, when would this happen? What is the user supposed to do with this information?

crates/wit/src/commands/pull.rs
```diff
@@ -205,7 +205,6 @@ impl PullCommand {
     }
 }
 
-#[derive(Debug)]
 struct PackagesState {
     // Packages currently present in the wit dir
     present: BTreeSet<PackageName>,
```
rylev, line 217: Nit: I'm not sure if `present` is the best term here. Perhaps `resolved`? 🤷‍♂️

crates/wit/src/commands/pull.rs
lann: Currently nothing enforces that the outer registry version (Warg release version or OCI tag) matches the inner WIT package version. This is a problem here because we only have the WIT version for the "already installed" set and can only use the registry version to resolve packages to pull.

I'm not sure how to resolve this yet, so the warning serves primarily to alert publishers that something is wrong.

rylev:
When I remove ghcr.io credentials by running `docker logout ghcr.io` then I end up with a different error:

```
error: failed to get registry credentials: Credential helper returned non-zero response code:
stdout:
credentials not found in native keychain


stderr:
```

And if I log back in with `docker login ghcr.io` then everything works as expected. 

So it seems we're not allowing anonymous access to the ghcr instance and we're not handling the error case of when the credentials are bad. 

I see you've opened https://github.com/lann/warg-loader/issues/5 as well which should hopefully address this.

lann:
I made 2 changes in warg-loader which should fix the above auth issues:
- Tweaked auth helper error handling to treat a non-zero exit code as "no credential found"; this is apparently what is expected :shrug: 
- Retry anonymous auth after credential failure, to catch the case where an expired token is used with a public registry

rylev (approved):
Not sure how I'd test an expired token, but this seems to have fixed the issue when no token is present. I think this is ready to merge!

lann:
> Not sure how I'd test an expired token

I tested by corrupting a token in `~/.docker/config.json` which seems to be the same error path as expiration.

Zelzahn:
If I'm understanding it correctly, this will not serve as a replacement for [wit-deps](https://github.com/bytecodealliance/wit-deps)?

lann:
> If I'm understanding it correctly, this will not serve as a replacement for [wit-deps](https://github.com/bytecodealliance/wit-deps)?

It might eventually. Part of the purpose of this tool is to be a proving ground for some of the more advanced import definition types now available in the component model which should be able to replace external dependency config files.

See "options for naming imports" here: https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions

rylev:
I believe this PR was closed by mistake by #270. I believe the correct issue to close was #268.

@peterhuene can you reopen this PR?

peterhuene:
Whoops, sorry about that!

calvinrp:
Closing this PR. The `wit` CLI functionality is in the process of moving over to https://github.com/bytecodealliance/wasm-pkg-tools

PR #270, opened 2024-04-12
Synchronize artifact componentization. 

### LINKED ISSUE #262: deps: bump wit-bindgen to `v0.22`
Issue for bumping `wit-bindgen` to `0.22.x` 

Ideally after the patch is released to address https://github.com/bytecodealliance/wit-bindgen/pull/891 

https://github.com/bytecodealliance/cargo-component/blob/658dc2ef060de5a7bd626d7f35fd786fb718c8fd/Cargo.toml#L105

(sorry for being high maintenance 🛠️, the deltas as messing up my builds)

peterhuene:
This PR makes componentization of build artifacts acquire the lock file in
an effort to prevent concurrent cargo-component processes that are building the
same workspace from interfering with one another.

Additionally, this commit changes the writing of the final build artifact so
that it writes to a temp file and then atomically persists the temp file at the
final path.

Fixes #265
Closes #262


src/lib.rs
```diff
@@ -983,12 +987,37 @@ fn componentize(
...
+            path = file.path().display()
+        )
+    })?;
+
+    file.as_file_mut().sync_all().with_context(|| {
+        format!(
+            "failed to sync output component `{path}`",
+            path = file.path().display()
+        )
+    })?;
```
sunfishcode, line 1011: Is there a specific need for a `sync_all` here? Compilers don't usually sync their output to disk, because it can be a slow operation on some filesystems, and it's only significant if there's eg. a power failure.

peterhuene: There's no specific need; I'll remove.

sunfishcode (approved):

Issue #257, opened 2024-03-12
Add support for `with` from `wit-bindgen`


peterhuene:
Implement support in `Cargo.toml` (i.e. `[package.metadata.component.bindings]`) for specifying how generated bindings should refer to already existing implementations.

See https://github.com/bytecodealliance/wit-bindgen/issues/840.

PR #273, opened 2024-04-17
add support for missing settings from wit-bindgen

sehnryr:
This adds support for the missing settings from wit-bindgen.
It is now a 1 to 1 copy of `Opts` fields (but `with`) from `wit-bindgen-rust` although not similarly sorted. (https://github.com/bytecodealliance/wit-bindgen/blob/90a1e547e1a4515a5769f5d157ffc519c5d851a3/crates/rust/src/lib.rs#L87-L189)
The `with` fields is represented here as a `HashMap<String, String>` instead of the `Vec<(String, String)>` from `Opts` since its TOML serialized representation is much more appropriate.

E.g. `Cargo.toml`
```toml
[package.metadata.component.bindings]
with = {foo: bar}
generate_unused_types = true
```

Closes: #257


peterhuene (approved):
Thanks for adding this!

PR #274, opened 2024-04-17
fix kebab case issue

### LINKED ISSUE #272: Hello world example fails when project name contains underscore
I haven't diagnosed this. But with a new cargo-component installed (v0.11.0), calling `cargo component new hello` builds a cargo directory within which `cargo component build` succeeds but if `cargo component new hello_world` is used, the build step then does not succeed.

```bash
try $ cargo component new hello_world
     Created binary (application) `hello_world` package
     Updated manifest of package `hello_world`
   Generated source file `src/main.rs`
 try $ cd hello_world
 hello_world $ cargo component build
  Generating bindings for hello_world (src/bindings.rs)
   Compiling wit-bindgen-rt v0.24.0
   Compiling bitflags v2.5.0
   Compiling hello_world v0.1.0 (/Users/frank/wasm/try/hello_world)
    Finished dev [unoptimized + debuginfo] target(s) in 0.50s
    Creating component target/wasm32-wasi/debug/hello_world.wasm
error: decoding custom section component-type:wit-bindgen:0.24.0:hello_world:encoded world

Caused by:
    export name `component:hello_world/hello_world` is not a valid extern name
    expected `/` after package name (at offset 0x26)
```

fibonacci1729:
Fixes #272

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -7,6 +7,7 @@ use crate::{
 };
 use anyhow::{bail, Context, Result};
 use cargo_component_core::registry::DecodedDependency;
+use heck::ToKebabCase;
 use indexmap::{IndexMap, IndexSet};
 use semver::Version;
 use std::{
@@ -431,7 +432,7 @@ impl<'a> BindingsGenerator<'a> {
 
     fn target_empty_world(resolution: &PackageDependencyResolution) -> (Resolve, WorldId) {
         let mut resolve = Resolve::default();
-        let name = resolution.metadata.name.clone();
+        let name = resolution.metadata.name.to_kebab_case();
         let pkg_name = PackageName {
             namespace: "component".to_string(),
             name: name.clone(),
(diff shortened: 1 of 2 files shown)


src/metadata.rs
```diff
@@ -378,7 +379,7 @@ impl ComponentMetadata {
         }
 
         Ok(Self {
-            name: package.name.clone(),
+            name: package.name.clone().to_kebab_case(),
```
peterhuene, line 382: The package name in the metadata should match the name from the cargo metadata, ideally.

Instead, I think this fix needs to be applied to https://github.com/bytecodealliance/cargo-component/blob/main/src/bindings.rs#L434 (replacing the `clone` call).

peterhuene (commented):
Additionally, could we add a build test for a bin project named `hello_world`?

src/metadata.rs
fibonacci1729: Thanks! Updated and added a build test.

peterhuene (approved):
Thanks for fixing this!

Issue #293, opened 2024-05-09
Compile error when installing cargo component


kajacx:
Recently, I got an error when I was installing `cargo component` in GitHub actions. The error is:

```
error[E0061]: this method takes 1 argument but 0 arguments were supplied
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-component-core-0.11.0/src/registry.rs:882:38
    |
882 |                 .load_package(client.get_warg_registry(), e.key())
    |                                      ^^^^^^^^^^^^^^^^^-- an argument of type `&str` is missing
    |
note: method defined here
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/warg-client-0.4.3/src/lib.rs:118:18
    |
118 |     pub async fn get_warg_registry(
    |                  ^^^^^^^^^^^^^^^^^
help: provide the argument
    |
882 |                 .load_package(client.get_warg_registry(/* &str */), e.key())
    |                                                       ~~~~~~~~~~~~
error[E0308]: mismatched types
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-component-core-0.11.0/src/registry.rs:882:31
    |
882 |                 .load_package(client.get_warg_registry(), e.key())
    |                  ------------ ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Option<&RegistryDomain>`, found future
    |                  |
    |                  arguments to this method are incorrect
    |
note: method defined here
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/warg-client-0.4.3/src/storage.rs:111:14
    |
111 |     async fn load_package(
    |              ^^^^^^^^^^^^
error[E0599]: no method named `upsert` found for struct `Arc<warg_client::Client<FileSystemRegistryStorage, FileSystemContentStorage, FileSystemNamespaceMapStorage>>` in the current scope
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cargo-component-core-0.11.0/src/registry.rs:594:32
    |
594 |                 (index, client.upsert(upserts.iter()).await)
    |                                ^^^^^^ method not found in `Arc<Client<FileSystemRegistryStorage, FileSystemContentStorage, FileSystemNamespaceMapStorage>>`
Some errors have detailed explanations: E0061, E0308, E0599.
For more information about an error, try `rustc --explain E0061`.
[...]

PR #288, opened 2024-05-06
Update to latest Warg release to support registry federation

calvinrp:
This is a small update to update to the latest [Warg release](https://github.com/bytecodealliance/registry/releases/tag/v0.4.2). Supports federation across registries.

PR #294, opened 2024-05-09
Lock warg-client and warg-protocol to 0.4.1

kajacx:
See https://github.com/bytecodealliance/cargo-component/issues/293


calvinrp:
#288 was merged and resolves this

Issue #299, opened 2024-05-20
Bindings are sometimes incorrectly generated type + `Borrow` suffix vs `&` type


calvinrp:
`cargo component new --lib --target calvinrp:scratch-example --registry https://calvinrp.wa.dev scratch-example`

generates:

```rust
#[allow(warnings)]
mod bindings;

use bindings::exports::calvinrp::scratch_example::service::{Guest, Key};

struct Component;

impl Guest for Component {
    fn sign(key: KeyBorrow) -> String {
        unimplemented!()
    }
}

bindings::export!(Component with_types_in bindings);
```

where
`sign` method's `key` arg should be `&Key` instead of `KeyBorrow`
and fails to compile since the `KeyBorrow` type does not exist

The WIT for the `calvinrp:scratch-example` is shown here:
```wit
package calvinrp:scratch-example@0.0.1;

/// An example world for the component to target.
world example {
    export service;
}

interface service {
  use other.{key};

  sign: func(key: borrow<key>) -> string;
}

interface other {
  resource key {}
}
```

calvinrp:
OK, figured it out. If the interface types are exported in the target world, then `wit-bindgen-rust` uses `SomeBorrow` instead of `&Some`. But `cargo-component` doesn't implement the same logic.

PR in the works

calvinrp:
#300 fixes

PR #300, opened 2024-05-21
Bug fixes for targeting a world with resource borrows

calvinrp:
Expanded test cases to cover issue #299 

Also, when specifying an exact semver requirement, adds that exact requirement to the generated `Cargo.toml`. There was mismatch before.

`wit-bindgen-rust` generates different bindings for `borrow<>` of resources, depending on whether that type is exported in the world or not. `&SomeType` vs `SomeTypeBorrow` in the generated bindings. Fixed the logic here to match that behavior. The revised test case should better cover future potential breakage.

If interested, here is some of the logic on the `wit-bindgen` side:
https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/rust/src/interface.rs#L2203-L2225


macovedj (approved):

src/generator.rs
```diff
@@ -416,28 +405,54 @@ impl<'a> UnimplementedFunction<'a> {
...
             }
-            TypeDefKind::Type(ty) => self.print_type(in_world, ty, trie, source)?,
-            TypeDefKind::Handle(Handle::Own(id)) => {
-                self.print_type_id(in_world, *id, trie, source)?
-            }
+            TypeDefKind::Type(ty) => self.print_type(ty, trie, source)?,
+            TypeDefKind::Handle(Handle::Own(id)) => self.print_type_id(*id, trie, source, false)?,
             TypeDefKind::Handle(Handle::Borrow(id)) => {
-                if in_world {
+                let type_info = &self.resolve.types[*id];
```
fibonacci1729, line 421: nit: Could we pull the body of this case out into a method called `is_export_resource` (or something)? Keeps it a little more readable IMO.

fibonacci1729 (approved):
Minor nit, but overall makes sense to me!

src/generator.rs
calvinrp: Yep, good call

calvinrp: Fixed, thanks

Issue #306, opened 2024-05-28
Update to `-S cli` instead of `-S common` for wasmtime commands


calvinrp:
See [wasmtime v20.0.0 release notes](https://github.com/bytecodealliance/wasmtime/blob/release-20.0.0/RELEASES.md#changed) and https://github.com/bytecodealliance/cargo-component/pull/305#discussion_r1616318078

PR #308, opened 2024-06-01
added wasmtime wasi args for `cargo component serve`

calvinrp:


fibonacci1729 (approved):

src/lib.rs
```diff
@@ -274,7 +274,7 @@ fn get_runner(serve: bool) -> Result<PathAndArgs> {
             (
                 PathAndArgs::new("wasmtime")
                     .args(if serve {
-                        vec!["serve"]
+                        vec!["serve", "-S", "common", "-S", "http"]
```
FrankReh, line 277: I think "common" is being deprecated. Too early to use "cli"?

calvinrp: Yeah, good catch. I have an issue to track making that change: https://github.com/bytecodealliance/cargo-component/issues/306

It was recently failing CI tests switching to "cli" instead of "common". Need to track down why it is still grabbing an older release of wasmtime.

PR #318, opened 2024-07-04
Stop generating the bindings file for projects without wit files and dependencies

### LINKED ISSUE #314: `cargo component build` always produces a `bindings.rs` file
I've been working on a number of projects which just directly pull the [`wasi`](https://docs.rs/wasi/latest/wasi/) crate, but don't do anything with a local `wit/` directory. This produces a `src/bindings.rs` file, despite not actually ever being used. This can be a little confusing for people trying to read the code, because there is currently always an unused but important-sounding file sitting in the project hierarchy.

Can we make generating this file conditional on, say, whether there is a local `wit/` directory present?

iawia002:
Such as the case described in #314.

cc @calvinrp 

close #314

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -73,26 +73,26 @@ impl<'a> BindingsGenerator<'a> {
     /// Returns a tuple of the bindings generator and a map of import names.
     pub fn new(
         resolution: &'a PackageDependencyResolution<'a>,
-    ) -> Result<(Self, HashMap<String, String>)> {
+    ) -> Result<Option<(Self, HashMap<String, String>)>> {
         let mut import_name_map = Default::default();
-        let (resolve, world, source_files) =
-            Self::create_target_world(resolution, &mut import_name_map).with_context(|| {
-                format!(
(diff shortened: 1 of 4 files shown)


calvinrp:
Do you recommend a good way to test locally?

iawia002:
Sure, a common scenario is using [wasi-rs](https://github.com/bytecodealliance/wasi-rs/blob/main/examples/cli-command.rs) directly to build wasm programs. In this case, we directly reference the [bindings generated by wasi-rs](https://github.com/bytecodealliance/wasi-rs/blob/main/src/bindings.rs) and do not have our wit files. In this scenario, an "empty" bindings file should not be generated.

Another scenario is directly using a wasm http client library such as [waki](https://github.com/wacker-dev/waki/tree/main/examples/client). We don't have our own wit files, but an ["empty" bindings file](https://github.com/wacker-dev/waki/blob/main/examples/client/src/bindings.rs) is generated.

You can directly use the code of the above two examples to test it locally.

calvinrp (approved):

PR #331, opened 2024-09-04
ref(*)!: Updates all code to use the new publish functionality in wkg

thomastaylor312:
This finishes replacing a direct dependency on warg with the newly released published functionality in wasm-pkg-tools. Please note that this removes some warg specific flags and it updates tests.

### DIFF
--- a/crates/wit/src/commands/publish.rs
+++ b/crates/wit/src/commands/publish.rs
@@ -1,9 +1,7 @@
 use anyhow::{Context, Result};
-use cargo_component_core::{cache_dir, command::CommonOptions, registry::find_url};
+use cargo_component_core::{cache_dir, command::CommonOptions};
 use clap::Args;
-use warg_crypto::signing::PrivateKey;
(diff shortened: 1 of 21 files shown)


crates/wit/src/commands/publish.rs
```diff
@@ -55,31 +48,14 @@ impl PublishCommand {
...
         publish_wit_package(
             PublishOptions {
                 config: &config,
                 config_path: &config_path,
-                warg_config: &warg_config,
                 pkg_config,
                 cache: file_cache,
-                url,
-                signing_key,
+                url: self.registry.as_ref(),
```
calvinrp, line 57: Nit but should we rename `url` field to `registry`, since it is [not a URL](https://github.com/bytecodealliance/wasm-pkg-tools/blob/main/crates/wasm-pkg-common/src/registry.rs#L11)

calvinrp (approved):

PR #346, opened 2024-10-11
Avoid overwriting `bindings.rs` if content is unchanged

tomasol:
Compare the contents of `bindings.rs` before overwriting. If no changes are detected, skip the overwrite to prevent unnecessary recompilation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -855,13 +855,14 @@ async fn generate_package_bindings(
             path = output_dir.display()
         )
     })?;
-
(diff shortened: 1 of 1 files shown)


sunfishcode:
This looks good, and I can confirm that this lets cargo avoid rebuilding entire projects all the time. Thanks!

sunfishcode (approved):

sunfishcode:
This is now released in cargo-component 0.20.

PR #370, opened 2025-01-21
Update the adapter binaries cargo-component uses

alexcrichton:
This pulls in a fix for bytecodealliance/wasmtime#10058


sunfishcode (approved):

alexcrichton:
The [CI failure](https://github.com/bytecodealliance/cargo-component/actions/runs/12897332856/job/35962214864?pr=370) looks like it's due to the fact that the updated adapter has WASI 0.2.3 as the version where it previously had WASI 0.2.0. The version of Wasmtime used in CI wasn't old enough to have the forward-compatibility fix (CI used 22, the fix was in 23). I've confirmed that Wasmtime 23, as expected, works well with the new adapter. 

I've gone ahead and pushed a commit that updates the version used in CI to Wasmtime 29.0.1 as well, the current latest release of Wasmtime.

PR #375, opened 2025-03-03
make `CargoArguments::parse_from` public

greenhat:
To allow using `cargo-component` as a library.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -411,7 +411,7 @@ impl CargoArguments {
     }
 
     /// Parses the arguments from an iterator.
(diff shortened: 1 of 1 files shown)


greenhat:
I checked the failing CI tests, and found that they are triggered by the new Rust edition. Seems to be they should be fixed by https://github.com/bytecodealliance/wit-bindgen/pull/1183.

greenhat:
Thank you! Will do.

pchickey:
I just got the 2024 edition stuff fixed on main.

greenhat:
@pchickey Thanks! Rebased.

PR #381, opened 2025-03-17
Cargo.toml: don't accept semver-compatible with parse_arg 0.1.4

pchickey:
parse_arg 0.1.6 does not build with stable.


sunfishcode (approved):

Cargo.toml
```diff
@@ -84,7 +84,7 @@ log = "0.4.21"
 oci-distribution = "0.11"
 owo-colors = "4.0.0"
 p256 = "0.13.2"
-parse_arg = "0.1.4"
+parse_arg = "=0.1.4"
```
Kixunil, line 87: Hey, this is quite broken solution because it will prevent updates to `parse_arg`. Projects should generally use `Cargo.lock` to pin their dependencies.

However, I'm looking into the cause of this issue.

pchickey: Our users install cargo-component as a binary using `cargo install cargo-component` so anyone who didnt pass `--locked` would fail to install. This crate is not intended to be used as a library.

Kixunil: Well, true, it's cargo's fault that `--locked` is not default and this kind of pinning does make sense for you now that I understand your usage. Regardless, it's fixed now.

Issue #378, opened 2025-03-14
Encode the git commit hash in the `revision` section


yoshuawuyts:
We should find a way to extract the commit hash from version control to populate the [`wasm_metadata::Revision`](https://docs.rs/wasm-metadata/latest/wasm_metadata/struct.Revision.html) custom section. This would address the following todo, introduced in https://github.com/bytecodealliance/cargo-component/pull/372#pullrequestreview-2683580039:

https://github.com/bytecodealliance/cargo-component/blob/bed7cf9296866eaed39ac0b2eccd807bcd7329d4/src/lib.rs#L1058-L1059

justingaffney:
This is fixed by #383 and #385

PR #383, opened 2025-03-22
Add Git commit hash to the revision field of component metadata

justingaffney:
Gets the commit hash from the component package's Git repository, if available, and writes it to the revision field of the component metadata. Resolves issue #378

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -531,6 +531,7 @@ dependencies = [
  "cargo_metadata",
  "clap",
  "futures",
+ "git2",
  "heck 0.5.0",
  "indexmap 2.7.0",
  "libc",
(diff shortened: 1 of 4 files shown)


pchickey (approved):

pchickey:
Unfortunately, landing this broke the publish-binaries CI on main, which isnt checked prior to the PR merging... I'll get to fixing that today

justingaffney:
Ah my bad, it built fine for me on Windows so I assumed it was good

pchickey:
Yep it was just a weird thing with cross-compiling for x86_64 from an arm64 mac. Presumably its solvable but it wasnt straightforward through brew, so it was easier to just get rid of the feature.

PR #385, opened 2025-03-24
git2 dep: set default-features = false

pchickey:
https://github.com/bytecodealliance/cargo-component/pull/383 broke the binary build https://github.com/bytecodealliance/cargo-component/actions/runs/14040229083, which is not checked prior to merge. The root cause is depending on OpenSSL in a cross-compiling context, which was not trivial to resolve. But there is also no reason to actually depend on OpenSSL, since our use of `git2` isn't related to connecting to servers. So, this PR disables all of the features of git2, because they are not needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1458,8 +1458,6 @@ dependencies = [
  "libc",
  "libgit2-sys",
  "log",
- "openssl-probe",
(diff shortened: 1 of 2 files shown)


yoshuawuyts (approved):
looks great, ty!

PR #400, opened 2025-05-13
feat: custom registries for targets in cargo-component new

vados-cosmonic:
This commit adds support for using custom registries with targets when performing `cargo component new`. Users can specify a registry URI like `oci://ghcr.io` along with a registry namespace prefix like `bytecodealliance` and get easy access to packages (e.g. `docs:adder`) that are provided therein.

**NOTE** This PR relies on [a related one in wasm-pkg-tools](https://github.com/bytecodealliance/wasm-pkg-tools/pull/161)
Until that PR is merged, the Cargo.toml in use here can't be updated to remove the pointer to my fork.

An example invocation:

```
cargo component new \
    --lib \
    --target docs:adder \
    --registry oci://ghcr.io \
    --registry-ns-prefix bytecodealliance \
    adder
```

While this invocation seems long, it is much shorter than configuring other tooling, and serves as a decent option for quickstarts IMO.

The one thing left to solve in this issue is how to enable cargo component to pull from namespace registry overrides from via it's own config (or alternatively, create a `wkg.toml` )

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,7 +111,7 @@ warg-server = "0.9.0"
 wasi-preview1-component-adapter-provider = "29.0.1"
 wasm-metadata = "0.227.0"
 wasm-pkg-client = "0.11.0"
-wasm-pkg-common = "0.11.0"
+wasm-pkg-common = { version = "0.11.0", features = ["oci_extras"] }
 wasmparser = "0.227.0"
 wasmprinter = "0.227.0"
 wat = "1.220.0"
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -352,15 +352,32 @@ impl NewCommand {
             }
         }
 
+        // Build table of dependencies
         let mut dependencies = Table::new();
-        component["dependencies"] = Item::Table(dependencies);
         if let Some(target) = &self.target {
             let pkg_ref = PackageRef::from_str(target)
                 .with_context(|| format!("converting [{target}] to package ref"))?;
             if let Some(meta) = config.pkg_config.package_registry_override(&pkg_ref) {
-                dependencies[target] = Dependency::Package(())
+                // TODO: create anon registry mapping
+                let registry = Some(String::from("anon_mapping"));
+                // TODO: properly determine version version
+                let version =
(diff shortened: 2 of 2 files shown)


vados-cosmonic:
It looks like adding `wkg.toml` into the created repository (i.e. adding the file to the output of `cargo component new`) is the most reasonable way forward, and *just works* with the upstream projects involved with minimal strain on the user's side.

Given that currently `wkg.toml` [only currently supports file path based overrides](https://github.com/bytecodealliance/wasm-pkg-tools/blob/4958c7397b1cfadd9b3c634491d49ef6465fea15/crates/wasm-pkg-core/src/config.rs#L23), I think the most reasonable path forward here is to add support there for namespace package overrides.

Cargo.toml
```diff
@@ -109,7 +110,13 @@ warg-protocol = "0.9.0"
...
 wasi-preview1-component-adapter-provider = "29.0.1"
 wasm-metadata = "0.227.0"
-wasm-pkg-client = "0.9.0"
+#wasm-pkg-client = "0.9.0"
+#wasm-pkg-common = "0.10.0"
+# TODO: remove once wasm-pkg-tools update to add
+wasm-pkg-client = { git = "https://github.com/vados-cosmonic/bytecodealliance_wasm-pkg-tools", rev = "f41aeda3cc6c7eee489d1b8c1316a645f8df2b0c" }
+wasm-pkg-common = { git = "https://github.com/vados-cosmonic/bytecodealliance_wasm-pkg-tools", rev = "f41aeda3cc6c7eee489d1b8c1316a645f8df2b0c", features = [
+  "oci_extras",
+] }
```
vados-cosmonic, line 119:
```suggestion
wasm-pkg-client = { git = "https://github.com/bytecodealliance/wasm-pkg-tools", rev = "5221bdca410563e2087d8adc36425361859e7263" }
wasm-pkg-common = { git = "https://github.com/bytecodealliance/wasm-pkg-tools", rev = "5221bdca410563e2087d8adc36425361859e7263", features = [
  "oci_extras",
] }
```

vados-cosmonic:
@macovedj @calvinrp not sure if we need an upstream release of `wasm-pkg-tools` here, but at the very least the new functionality is in upstream there!

